PR #31, opened 2015-02-20
Add example script that uses xterm

raboof:
rather than osascript and Terminal, nice for linux users :)


raboof:
Makes to try and auto-detect that, see #34

PR #113, opened 2015-09-15
Implement a LWWRegister CmRDT

ianclegg:
This is an initial implementation of a LWWRegister using wall clocks (system time) to resolve conflicts. The register is 'nullable', not sure whether this is a good idea, but the MVRegister seemed to be too. Im new to eventuate, and open to feedback

### DIFF
--- a/src/main/scala/com/rbmhtechnology/eventuate/crdt/LWWRegister.scala
+++ b/src/main/scala/com/rbmhtechnology/eventuate/crdt/LWWRegister.scala
@@ -1,0 +1,106 @@
+/*
+ * Copyright (C) 2015 Red Bull Media House GmbH <http://www.redbullmediahouse.com> - all rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rbmhtechnology.eventuate.crdt
+
+import akka.actor.ActorRef
+import akka.actor.ActorSystem
+import com.rbmhtechnology.eventuate.{Versioned, DurableEvent, VectorTime}
+
+import scala.concurrent.Future
+
+/**
+ * A LWWVersioned value.
+ *
+ * @param value The value.
+ * @param vectorTimestamp Timestamp of the event that caused this version.
+ * @param systemTimestamp Wall Clock
+ * @param emitterId Creator of the event that caused this version.
+ */
+case class LWWVersioned[A](value: A, vectorTimestamp: VectorTime, systemTimestamp: Long = 0L, emitterId: String = "")
+
+/**
+ * Replicated Last Writer Wins Register. Stores a single value and timestamp at which it was set. The value is updated
+ * only if the vector clock is later, if the vector clocks are concurrent the register compares the system time,
+ * if the values are still concurrent the source with the greatest emitterId wins
+ * Note that this relies on synchronized system clocks. [[LWWRegister]] should only be used when the choice of
+ * value is not important for concurrent updates occurring within the clock skew.
+ *
+ * @param versionedValue Initially empty.
+ *
+ * @see [[http://hal.upmc.fr/docs/00/55/55/88/PDF/techreport.pdf A comprehensive study of Convergent and Commutative Replicated Data Types]]
+ */
+case class LWWRegister[A](versionedValue: Option[LWWVersioned[A]] = None) {
+
+  def value: Option[A] =
+    versionedValue.map(_.value)
+
+  def set(v: A, vt : VectorTime, st : Long, emitterId : String): LWWRegister[A] = {
+    val nvv = LWWVersioned(v, vt, st, emitterId)
+    val result = versionedValue match {
+      case None =>
+        Some(nvv)
+      case Some(vv) if nvv.vectorTimestamp > vv.vectorTimestamp =>
+        Some(nvv)
+      case Some(vv) if nvv.systemTimestamp > vv.systemTimestamp =>
+        Some(nvv)
+      case Some(vv) if nvv.systemTimestamp == vv.systemTimestamp && nvv.emitterId > vv.emitterId =>
+        Some(nvv)
+      case _ =>
+        versionedValue
+    }
+    copy(result)
+  }
+}
+
+object LWWRegister {
+  def apply[A]: LWWRegister[A] =
+    new LWWRegister[A]()
+
+  implicit def LWWRegisterServiceOps[A] = new CRDTServiceOps[LWWRegister[A], Option[A]] {
+    override def zero: LWWRegister[A] =
+      LWWRegister.apply[A]()
+
+    override def value(crdt: LWWRegister[A]): Option[A] =
+      crdt.value
+
+    override def precondition: Boolean =
+      false
+
+    override def update(crdt: LWWRegister[A], operation: Any, event: DurableEvent): LWWRegister[A] = operation match {
+      case UpdateOp(value) => crdt.set(value.asInstanceOf[A], event.vectorTimestamp, event.systemTimestamp, event.emitterId)
+    }
+  }
+}
+
+/**
+ * Replicated [[LWWRegister]] CRDT service.
+ *
+ * @param serviceId Unique id of this service.
+ * @param log Event log.
+ * @tparam A LWWRegister value type.
+ */
+class LWWRegisterService[A](val serviceId: String, val log: ActorRef)(implicit system: ActorSystem, val ops: CRDTServiceOps[LWWRegister[A], A])
+  extends CRDTService[LWWRegister[A], A] {
+
+  /**
+   * Updates the value
+   */
+  def update(id: String, value: A): Future[A] =
+    op(id, SetOp(value))
+
+  start()
+}
--- a/src/test/scala/com/rbmhtechnology/eventuate/crdt/CRDTSpec.scala
+++ b/src/test/scala/com/rbmhtechnology/eventuate/crdt/CRDTSpec.scala
@@ -22,6 +22,7 @@ import org.scalatest._
 
 class CRDTSpec extends WordSpec with Matchers with BeforeAndAfterEach {
   val mvReg = MVRegister[Int]()
+  val lwwReg = LWWRegister[Int]()
   val orSet = ORSet[Int]()
 
   def vectorTime(t1: Long, t2: Long): VectorTime =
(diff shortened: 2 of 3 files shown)


ianclegg:
Build break with the Cassandra integration tests related to connections. The tests passed locally and the commit was not related to cassandra. Any ideas?

All host(s) tried for query failed (tried: /127.0.0.1:9142 (com.datastax.driver.core.TransportException: [/127.0.0.1:9142] Cannot connect)

krasserm:
Thanks a lot for your PR, I'll take a closer look tomorrow. Regarding the test failure, see [my comment](https://gitter.im/RBMHTechnology/eventuate?at=55f84283c588a0de6ede9ea5) on Gitter.

krasserm:
I just re-started the build and the [compile errors](https://travis-ci.org/RBMHTechnology/eventuate/builds/80457010#L411) are now because of [my recent changes](https://github.com/RBMHTechnology/eventuate/commit/969bdd67a70da302c75f6be607a9782d54cd9b6c#diff-89829be350570a9e7e7ba5c89128f182R63) in master. Please rebase your PR and go a `git push -f`. Thanks!

ianclegg:
Your quite right on all points and the vector clock is nice addition if clocks are synchronized, i'll take a look at this shortly.

krasserm:
@ianclegg are you still planning to work on it?

ianclegg:
Yes, unless it is blocking you - I'll pick this up again this evening

krasserm:
@ianclegg not at all, was just curious :smiley:. Great that you continue working on it.

ianclegg:
@krasserm It seemed to make sense to reuse Versioned to encapsulate the value and the vector clock. Is it fair to use the existing 'creator' property of Versioned to store the emitter id too. I guess what I would like to do is have both the vector clock and the wall clock in Versioned

something like this:
```scala
case class Versioned[A](value: A, vectorTimestamp: VectorTime, timestamp: Long, creator: String = "")
```

Then I can refactor the systemTimestamp out

```scala
case class LWWRegister[A](versionedValue: Versioned[Option[A]] = Versioned(None, VectorTime(), ""),
                          systemTimestamp : Long = 0L) {

  def value: A =
    versionedValue.value.get

  def set(newValue: A, vectorTimestamp : VectorTime, newTimestamp : Long, emitterId : String): LWWRegister[A] = {
        // First, compare the vector timestamps
    if ((vectorTimestamp > versionedValue.updateTimestamp) ||
        // If the vector timestamps are the same, compare the system timestamps
        (vectorTimestamp == versionedValue.updateTimestamp) &&
          (newTimestamp > systemTimestamp) ||
        // If the system timestamps are the same, choose the value with the lowest emitter
        (newTimestamp == systemTimestamp) &&
        (emitterId > versionedValue.creator)
    )
        copy(versionedValue.copy(Some(newValue), vectorTimestamp, emitterId), newTimestamp)
    else
      copy(versionedValue, systemTimestamp)
  }
}
```
Is this along the lines of what you were thinking?

Also,  there are cases where it might be desirable to ignore the vector clock and use the wall clock. For example, in financial systems where a trader is only ever interested in the latest price - regardless of the ordering of previous events.

krasserm:
> Is it fair to use the existing 'creator' property of Versioned to store the emitter id too

Yes.

> have both the vector clock and the wall clock in Versioned

I'm a bit hesitant to a add a wall clock timestamp to `Versioned`. There are many other usages of `Versioned` in Eventuate (and the example application) which don't need that timestamp. I'd rather propose a field of type `Option[(Versioned[A], Long)]` to store what you need. A later refactoring can still be done when needed.

> `// If the vector timestamps are the same`

This will never be the case. If `vectorTimestamp > versionedValue.updateTimestamp` is `false` then 
`vectorTimestamp conc versionedValue.updateTimestamp` will be `true`. Vector timestamps are unique [unique event identifiers](https://github.com/RBMHTechnology/eventuate/blob/v-0.3/src/main/scala/com/rbmhtechnology/eventuate/DurableEvent.scala#L62).

ianclegg:
> There are many other usages of Versioned in Eventuate (and the example application) which don't need that timestamp

agreed

> I'd rather propose a field of type ```Option[(Versioned[A], Long)]``` to store what you need

agreed. The reasoning behind ```Versioned[Option[A]]``` was to enable the register to be cleared (ie. set to None) by a future operation - but ```Option[(Versioned[A], Long)]``` is closer to the behavior of the existing MVRegister

krasserm:
@ianclegg sorry for the delay, haven't been notified by Github that you've made new commits. Will make a detailed review tomorrow. 

Do you mind also updating the sentence 

> Eventuate currently implements 3 out of 12 operation-based CRDTs specified in the paper. These are Counter, MV-Register and OR-Set.

in the CRDT section of the [user guide](https://raw.githubusercontent.com/RBMHTechnology/eventuate/r-0.3/src/sphinx/user-guide.rst)? Should you want to build the documentation locally, [here are the instructions](http://rbmhtechnology.github.io/eventuate/developers.html#documentation).

src/main/scala/com/rbmhtechnology/eventuate/crdt/LWWRegister.scala
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * A LWWVersioned value.
+ *
+ * @param value The value.
+ * @param vectorTimestamp Timestamp of the event that caused this version.
+ * @param wallTimestamp Wall Clock
+ * @param creator Creator of the event that caused this version.
+ */
+case class LWWVersioned[A](value: A, vectorTimestamp: VectorTime, wallTimestamp: Long = 0L, creator: String = "")
```
krasserm: - Rename `wallTimestamp` to `systemTimestamp` to be consistent with naming used elsewhere?
- Maybe `creator` can also be renamed to `emitterId` since `LWWVersioned` is only used here?

src/main/scala/com/rbmhtechnology/eventuate/crdt/LWWRegister.scala
```diff
@@ -0,0 +1,104 @@
...
+ * value is not important for concurrent updates occurring within the clock skew.
+ *
+ * @param versionedValue Initially empty.
+ *
+ * @see [[http://hal.upmc.fr/docs/00/55/55/88/PDF/techreport.pdf A comprehensive study of Convergent and Commutative Replicated Data Types]]
+ */
+case class LWWRegister[A](versionedValue: Option[LWWVersioned[A]] = None) {
+
+  def value: Option[A] =
+    if(versionedValue.isDefined) Some(versionedValue.get.value) else None
```
krasserm: `versionedValue.map(_.value)`

src/main/scala/com/rbmhtechnology/eventuate/crdt/LWWRegister.scala
```diff
@@ -0,0 +1,104 @@
...
+          (wallTimestamp > versionedValue.get.wallTimestamp)) ||
+        // If we still have a tie, choose the version with greatest emitter id
+        (wallTimestamp == versionedValue.get.wallTimestamp) &&
+          (emitterId > versionedValue.get.creator)))
+    )
+      copy(Some(LWWVersioned(newValue, vectorTimestamp, wallTimestamp, emitterId)))
+    else
+      copy(versionedValue)
+  }
+}
```
krasserm: I find that very difficult to read. What do you think about

```scala
case class LWWRegister[A](versionedValue: Option[LWWVersioned[A]] = None) {

  def value: Option[A] =
    versionedValue.map(_.value)

  def set(nvv: LWWVersioned[A]): LWWRegister[A] = {
    val result = versionedValue match {
      case None => 
        Some(nvv)
      case Some(vv) if nvv.vectorTimestamp > vv.vectorTimestamp => 
        Some(nvv)
      case Some(vv) if nvv.systemTimestamp > vv.systemTimestamp =>
        Some(nvv)
      case Some(vv) if nvv.systemTimestamp == vv.systemTimestamp && nvv.emitterId > vv.emitterId =>
        Some(nvv)
      case _ => 
        versionedValue
    }
    copy(result)
  }
}
```

together with a corresponding change in `LWWRegisterServiceOps`?

src/main/scala/com/rbmhtechnology/eventuate/crdt/LWWRegister.scala
```diff
@@ -0,0 +1,104 @@
...
+      LWWRegister.apply[A]()
+
+    override def value(crdt: LWWRegister[A]): Option[A] =
+      crdt.value
+
+    override def precondition: Boolean =
+      false
+
+    override def update(crdt: LWWRegister[A], operation: Any, event: DurableEvent): LWWRegister[A] = operation match {
+      case UpdateOp(value) => crdt.set(value.asInstanceOf[A], event.vectorTimestamp, event.systemTimestamp, event.emitterId)
```
krasserm: See previous comment.

src/main/scala/com/rbmhtechnology/eventuate/crdt/LWWRegister.scala
```diff
@@ -0,0 +1,104 @@
...
+ * @param creator Creator of the event that caused this version.
+ */
+case class LWWVersioned[A](value: A, vectorTimestamp: VectorTime, wallTimestamp: Long = 0L, creator: String = "")
+
+/**
+ * Replicated Last Writer Wins Register. Stores a single value and timestamp at which it was set. The value is updated
+ * only if the timestamp of the new value is greater than the timestamp of the current value.
+ * Note that this relies on synchronized system clocks. [[LWWRegister]] should only be used when the choice of
+ * value is not important for concurrent updates occurring within the clock skew.
+ *
```
krasserm: Should this description also mention usage of vector time and emitter id?

src/test/scala/com/rbmhtechnology/eventuate/crdt/CRDTSpec.scala
```diff
@@ -63,6 +64,41 @@ class CRDTSpec extends WordSpec with Matchers with BeforeAndAfterEach {
...
+      lwwReg
+        .set(1, vectorTime(1, 0), 1, "emitter-1")
+        .set(2, vectorTime(2, 0), 0, "emitter-2")
+        .value should be(Some(2))
+    }
+    "ignore a new value if it was set before the current value according to the vector clock" in {
+      lwwReg
+        .set(1, vectorTime(2, 0), 0, "emitter-1")
+        .set(2, vectorTime(1, 0), 1, "emitter-2")
+        .value should be(Some(1))
```
krasserm: This test should be removed as Eventuate will never deliver events in wrong causal order.

krasserm:
@ianclegg how do you plan to continue with this PR?

ianclegg:
@krasserm I just got back from vacation. I owe you a beer, the match expression is much more readable! I'll push an update later today

ianclegg:
@krasserm 
Versioned is not used on the interface of the the existing ORSet and MVRegister, so clients can use the CRDTs without knowing about it.

```scala
mvReg.set(1, vectorTime(1, 0))
orSet.add(1, vectorTime(1, 0))
```

Changing set() method in LWWRegister to take a LWWVersioned[A] means clients now need to first build a LWWVersioned which makes it a little more difficult to use. The match expression is nice, perhaps if I build the LWWVersioned within set() function.

```scala
def set(v: A, vt : VectorTime, st : Long, emitterId : String): LWWRegister[A] = {
    val nvv = LWWVersioned(v, vt, st, emitterId)
    val result = versionedValue match {
```

so clients can use LWWRegister much like MVRegister and ORSet?

```scala
lwwReg.set(1, vectorTime(1, 0), 0, "source-1")
```

krasserm:
Great, looks good to me now. Thanks @ianclegg. Please squash your commits and rebase on latest origin/master, then I'll merge.

ianclegg:
@krasserm rebased and squashed

Issue #99, opened 2015-08-12
Creation of ReplicationEndpoints with illegal ids should fail


volkerstampa:
The ids of `ReplicationEndpoint`s are used as part of the log-id which in turn is used by the event-log-backends to _name_ their storages. For example the cassandra-backend names the table that is used to store the events after this id. Different backends have different constraints regarding these names and thus regarding the log-id respectively the ReplicationEndpoint-id.

I propose that the creation of an `ReplicationEndpoint` with an _illegal_ id should fail immediately. Another option would be to implement proper escaping of these ids for each event-log-backend.

krasserm:
Covered by #124 (and escaping done by `LeveldbEventLog`).

PR #124, opened 2015-10-23
check cassandra event log id for valid characters

kongo2002:
Hi,

I just implemented a basic check/validation for the `id` of the `CassandraEventLog`. Since that `id` is used as part of the Cassandra tables that are used, we better throw an appropriate exception than failing with a rather *cryptic* SQL exception.
This one is closely related to #99 I guess.

Cheers,
Gregor


krasserm:
@kongo2002 awesome, thanks you very much. Appreciate your contributions :smiley:

PR #131, opened 2015-10-30
Instantiante Written just once

ermanito:
During some load testing with high amount of actors we faced many "Written" objects in the Heap. Commit is no big deal. We are just creating less Written instance.


krasserm:
Thanks for the contribution @ermanito - very useful enhancement!

PR #138, opened 2015-11-16
Cassandra: do not wait for system termination

kongo2002:
Hi there,

in case of a failure in the cassandra extension initialization we rather throw an exception than waiting (indefinitely) for the actor system to shut down.

As of now a process that fails to connect to the cassandra cluster will indefinitely wait until being killed by `kill -9`.

Cheers,
Gregor


krasserm:
Your changes make absolutely sense @kongo2002, thanks for the pull request. I'm anyway not happy with the `Cassandra` extension based session management and plan to replace it with a better solution in the future. Anyway, I'll merge your changes as they are a significant improvement on the current situation.

PR #195, opened 2016-01-18
Fix save snapshot failures when using PersistOnEvent

### LINKED ISSUE #194: Save snapshot can fail when using persistOnEvent
`persistOnEvent` is based on the reliable delivery of `PersistOnEventRequest`s to self. As such, these requests are added to the `_unconfirmed` delivery buffer of `ConfirmedDelivery` which is also included into snapshots. The problem is that `PersistOnEventRequest` contains completion handlers that are usually not serializable and may also close of current actor state. For this and other reasons, support for `persistOnEvent` handlers should be removed:

- Invoking the completion handler with `Success(event)` is equivalent to handling that event with the condition `if (!recovering)` directly in the event handler.
- Invoking the completion handler with `Failure(...)` anyway requires the application to throw an exception to restart the actor. This should be automated by internally throwing a `PersistOnEventException` and react on that exception, if needed, by overriding `preRestart`.

The above resolutions would make it possible to exclude handlers from `PersistOnEventRequest`s and make the requests serializable. Furthermore, a fix for this ticket should also include a protobuf serializer for `PersistOnEventRequest` (which delegates to a custom serializer for the payloads).

krasserm:
- fixes #194
- No user-defined persistOnEvent handlers
- Persisted events must be handler in event handler
- Persistence failures throw PersistOnEventException

- Re-implementation of PersistOnEvent
- Composition with ConfirmedDelivery possible
- redeliverUnconfirmed allowed in context of PersistOnEvent

### DIFF
--- a/src/it/scala/com/rbmhtechnology/eventuate/PersistOnEventIntegrationSpec.scala
+++ b/src/it/scala/com/rbmhtechnology/eventuate/PersistOnEventIntegrationSpec.scala
@@ -19,6 +19,7 @@ package com.rbmhtechnology.eventuate
 import akka.actor._
 import akka.testkit._
 
+import com.rbmhtechnology.eventuate.EventsourcedView.Handler
 import com.rbmhtechnology.eventuate.log.EventLogLifecycleCassandra
 import com.rbmhtechnology.eventuate.log.EventLogLifecycleLeveldb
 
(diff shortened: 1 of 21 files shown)


volkerstampa:
LGTM

PR #199, opened 2016-01-19
Use better default for event batch sizes

### LINKED ISSUE #181: Use better batch size default values
Chaos tests have shown that some of the batch size values are actually too high in `reference.conf` (breaking index updates, for example). Use batch size configurations from chaos tests as better defaults. This ticket is a preliminary enhancement until #166 is implemented.

krasserm:
- closes #181


volkerstampa:
+1

PR #210, opened 2016-02-11
Fix rejected execution of PersistOnEventRequests loaded from a snapshot

### LINKED ISSUE #209: PersistOnEventRequest not properly recovered from snapshot
`PersistOnEventRequest`s saved in a snapshot are not re-executed because of their old `instanceId`. `PersistOnEventRequest` recovery must replace the stored `instanceId` with the current one.

krasserm:
- fixes #209


volkerstampa:
LGTM

PR #212, opened 2016-02-11
Support custom serializers with string-manifest

### LINKED ISSUE #207: Support SerializerWithStringManifest when serializing event payload and replication filters
[Custom Serialization](http://rbmhtechnology.github.io/eventuate/reference/event-sourcing.html#custom-serialization) of events, snapshots and replication filters should support [serializers with string manifest](http://doc.akka.io/docs/akka/2.4.1/scala/serialization.html#Serializer_with_String_Manifest) introduced with akka 2.4.

volkerstampa:
When serializing a DurableEvent's or snapshots payload or
a custom ReplicationFilter include a string-manifest,
if the custom serializer is a SerializerWithStringManifest
and use corresponding deserialize-method.

closes #207


krasserm:
LGTM, great work!

PR #215, opened 2016-02-16
Defer deletion of events required for internal state recovery

### LINKED ISSUE #214: LevelDbEventLog cannot restore correct EventLogClock after event deletion
LevelDbEventLog stores _periodically_ a snapshot of the current EventLogClock and restores its state at startup from this snapshot and following events. In case of physical deletion of these events the state cannot properly be restored.

volkerstampa:
- Do not delete events after LevelDbEventLog's EventLogClock snapshot
  as they are required for EventLogClock-recovery at startup
- closes #214


krasserm:
LGTM, good catch, thanks for fixing!

Issue #208, opened 2016-02-10
Database replication proof-of-concept


krasserm:
A proof-of-concept for asynchronously replicating the database of an existing Spring/RDBMS application across multiple locations including conflict detection and resolution.

PR #228, opened 2016-02-29
Java-based version of database replication POC via inbound and outbound CDC

cstub:
- see #208


krasserm:
Awesome work @cstub. I'm just curious about the serializer configs. Will merge when answered.

PR #238, opened 2016-03-15
Java version of Query-DB example application

cstub:
Java-based version of query-db example application.


krasserm:
LGTM, great work!

PR #239, opened 2016-03-19
Avoid LevelDB shutdown while iterators are in use

### LINKED ISSUE #234: LevelDB event log crashes JVM on shutdown frequently
There's already PR #231 which will fix this, I'm just adding this here for reference. **Update:** this is now fixed by #239.

This is because leveldb iterators may still be in use when native data structures are decommissioned by the `leveldb.close()` call (despite the current iterator tracking code).

krasserm:
- Use iterators in child actors and rely on Akka to first stop these child actors before stopping the parent event log actor that shuts down LevelDB.
- fixes #234


volkerstampa:
LGTM

PR #241, opened 2016-03-23
Wip 240 no index update on cassandra log init

krasserm:


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/log/EventLog.scala
```diff
@@ -98,33 +106,43 @@ case class BatchReadResult(events: Seq[DurableEvent], to: Long) extends DurableE
...
  */
 private class PhysicalDeletionNotSupportedException extends UnsupportedOperationException
 
 /**
  * Event log storage provider interface (SPI).
+ *
+ * @tparam A Event log state type.
  */
-trait EventLogSPI { this: Actor =>
+trait EventLogSPI[A] { this: Actor =>
```
volkerstampa: Shouldnt we add here `<: EventLogState` as well?

eventuate-log-cassandra/src/main/scala/com/rbmhtechnology/eventuate/log/cassandra/CassandraEventLog.scala
```diff
@@ -84,11 +86,27 @@ class CassandraEventLog(id: String) extends EventLog(id) {
...
 
+  private var index: ActorRef = null
   private var indexSequenceNr: Long = 0L
   private var updateCount: Long = 0L
 
+  override def recoverState: Future[CassandraEventLogState] = {
+    import services.readDispatcher
+    for {
+      dm <- deletedToStore.readDeletedToAsync.map(DeletionMetadata(_, Set.empty))
+      sc <- indexStore.readEventLogClockAsync
```
volkerstampa: Could it make sense to call this method `readEventLogClockSnapshotAsync`?

volkerstampa:
LGTM

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/log/EventLog.scala
krasserm: That's actually only a constraint of the abstract `EventLog` (from which all implementations extend).

eventuate-log-cassandra/src/main/scala/com/rbmhtechnology/eventuate/log/cassandra/CassandraEventLog.scala
krasserm: Good idea, will do.

PR #246, opened 2016-04-11
Store deliveryId as event optional meta-datum in DurableEvent

### LINKED ISSUE #244: Do not require storage of deliveryId in application-specific events
Application-specific events used as confirmation events in context of `ConfirmedDelivery` shouldn't care about `deliveryId` storage (as extra field, for example). Instead, the following re-design of `ConfirmedDelivery` is proposed:

- optional `deliveryId` field in `DurableEvent`
- storage of confirmation events is done with `persistConfirmation`, a specialization of `persist` that has an extra `deliveryId` parameter. The `persistConfirmation` method is defined on `ConfirmedDelivery`.
- event handlers don't have to call `confirm` any more. If a `DurableEvent` with a defined `deliveryId` is received, the corresponding reliable message is automatically removed from the internal delivery queue (if the event's `emitterId` equals the actor's `id`).

krasserm:
- breaking change of confirmed delivery API
- closes #244

### DIFF
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ConfirmedDelivery.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ConfirmedDelivery.scala
@@ -27,16 +27,18 @@ object ConfirmedDelivery {
 }
 
 /**
- * Supports the reliable delivery of messages to destinations by enabling applications
- * to redeliver messages until they are confirmed by their destinations. Both, message
- * delivery and confirmation processing must be done within an [[EventsourcedActor]]'s
- * event handler. New messages are delivered by calling `deliver`. When the destination
- * replies with a confirmation message, the [[EventsourcedActor]] must generate an event
- * for which the handler calls `confirm`. Until confirmation, delivered messages are
- * tracked as ''unconfirmed'' messages. Unconfirmed messages can be redelivered by calling
- * `redeliverUnconfirmed`. This is usually done within a command handler by processing
- * scheduler messages. Redelivery occurs automatically when the [[EventsourcedActor]]
- * successfully recovered after initial start or a re-start.
+ * Supports the reliable delivery of messages to destinations by enabling applications to
+ * redeliver messages until they are confirmed by their destinations. The correlation
+ * identifier between a reliable message and its confirmation message is an
+ * application-defined `deliveryId`. Reliable messages are delivered by calling `deliver` in
+ * an [[EventsourcedActor]]'s event handler. When the destination replies with a confirmation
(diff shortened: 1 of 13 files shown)


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ConfirmedDelivery.scala
```diff
@@ -62,12 +71,6 @@ trait ConfirmedDelivery extends EventsourcedActor {
   }
 
   /**
-   * Confirms the message delivery identified by `deliveryId`.
-   */
-  def confirm(deliveryId: String): Unit =
```
volkerstampa: The scala-doc on `deliver` still mentions `confirm` instead of `persistConfirmation`

src/sphinx/reference/event-sourcing.rst
```diff
@@ -269,11 +269,11 @@ Reliable, event-based remote communication between event-sourced actors should b
...
+The ``destination`` confirms the delivery of the message by sending a ``Confirmation`` reply to the event-sourced actor from which it generates a ``ConfirmationEvent``. The actor uses the ``persistConfirmation`` method to persist the confirmation event together with the delivery id. After successful persistence of the confirmation event, the corresponding reliable message is removed from the internal buffer of unconfirmed messages.
 
-When the actor is re-started, unconfirmed ``ReliableMessage``\ s are automatically re-delivered to their ``destination``\ s. The example actor additionally schedules ``redeliverUnconfirmed`` calls to periodically re-deliver unconfirmed messages. This is done within the actor’s command handler.
+When the actor is re-started, unconfirmed reliable messages are automatically re-delivered to their ``destination``\ s. The example actor additionally schedules ``redeliverUnconfirmed`` calls to periodically re-deliver unconfirmed messages. This is done within the actor’s command handler.
```
volkerstampa: I am not sure, but I wonder, if we should explicitly mention that for this the id of the `EventsourcedActor` must not change, otherwise a replay will produce no confirmations.

volkerstampa:
LGTM

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ConfirmedDelivery.scala
krasserm: Good catch, thanks!

src/sphinx/reference/event-sourcing.rst
krasserm: I think it's better to mention elsewhere that the `id` of an `EventsourcedActor` must never change. Mentioning it here in this context might create the impression that it is allowed elsewhere. I'll cover that in a separate documentation ticket.

PR #247, opened 2016-04-11
Upper bound on source log scan progress

### LINKED ISSUE #236: Upper bound on source log scan progress per replication-read request
At the moment, replication read progress in the source log is only limited by the number of events that pass a given replication filter. However, a filter that drops a large number of events can cause a replication read request to timeout if the log scanning time exceeds the request timeout. This causes a replicator to re-send the requests starting another source log scan which will timeout again. Consequently, event replication not only isn't able to make progress but also overloads the replication source with log scans. An upper bound on scan progress mitigates this problem.

krasserm:
- can be configured with eventuate.log.replication.remote-scan-size
- closes #236

### DIFF
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
@@ -280,9 +280,9 @@ private class RecoveryActor(endpointId: String, link: RecoveryLink) extends Acto
   def recoveringMetadata: Receive = {
     case r: ReplicationRead if r.fromSequenceNr > link.localClock.sequenceNr + 1L =>
       println(s"[recovery of ${endpointId}] Trigger update of inconsistent replication progress at ${link.remoteLogId} ...")
-      sender() ! ReplicationReadSuccess(Seq(), link.localClock.sequenceNr, link.remoteLogId, link.localClock.versionVector)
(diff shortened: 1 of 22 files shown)


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
```diff
@@ -280,9 +280,9 @@ private class RecoveryActor(endpointId: String, link: RecoveryLink) extends Acto
   def recoveringMetadata: Receive = {
     case r: ReplicationRead if r.fromSequenceNr > link.localClock.sequenceNr + 1L =>
       println(s"[recovery of ${endpointId}] Trigger update of inconsistent replication progress at ${link.remoteLogId} ...")
-      sender() ! ReplicationReadSuccess(Seq(), link.localClock.sequenceNr, link.remoteLogId, link.localClock.versionVector)
+      sender() ! ReplicationReadSuccess(Seq(), link.localClock.sequenceNr, link.remoteLogId, link.localClock.versionVector, hasMore = false)
     case r: ReplicationRead =>
-      sender() ! ReplicationReadSuccess(Seq(), r.fromSequenceNr - 1L, link.remoteLogId, link.localClock.versionVector)
+      sender() ! ReplicationReadSuccess(Seq(), r.fromSequenceNr - 1L, link.remoteLogId, link.localClock.versionVector, hasMore = false)
```
volkerstampa: Coulnt we say `hasMore=true` here to _speed up_ recovery a little bit?

volkerstampa:
LGTM

PR #251, opened 2016-04-12
Fix replication performance regression

### LINKED ISSUE #250: Event replication performance regression
Seen in `BasicReplicationThroughputSpec` after application of #247

krasserm:
- fixes #250


volkerstampa:
+1

PR #252, opened 2016-04-13
Wip 222 replication failure detail reporting

krasserm:


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/log/EventLog.scala
```diff
@@ -395,7 +395,7 @@ abstract class EventLog[A <: EventLogState](id: String) extends Actor with Event
       remoteReplicationProgress += targetLogId -> (0L max from - 1)
       replicationRead(from, clock.sequenceNr, max, scanLimit, evt => evt.replicable(currentTargetVersionVector, filter)) onComplete {
         case Success(r) => self.tell(ReplicationReadSuccess(r.events, from, r.to, targetLogId, null), sdr)
-        case Failure(e) => self.tell(ReplicationReadFailure(e.getMessage, targetLogId), sdr)
+        case Failure(e) => self.tell(ReplicationReadFailure(e, targetLogId), sdr)
```
volkerstampa: I wonder if this is a bit risky?!? The exception can very well be backend specific, right? It gets serialized with standard java serialization and it might not be on the classpath of the receiving application which would make deserialization impossible, right?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -548,27 +550,44 @@ private class Replicator(target: ReplicationTarget, source: ReplicationSource, f
...
+  private val failureDetectionLimitNanos = failureDetectionLimit.toNanos
+  private var lastReportedAvailability: Long = 0L
 
   def receive = {
-    case Tick =>
-      val currentTime = System.currentTimeMillis
-      val lastInterval = currentTime - lastTick
-      if (lastInterval >= failureDetectionLimitMillis) {
+    case AvailabilityDetected =>
+      val currentTime = System.nanoTime()
```
volkerstampa: Why is `currentTimeMillis` not enough?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -548,27 +550,44 @@ private class Replicator(target: ReplicationTarget, source: ReplicationSource, f
...
-        lastTick = currentTime
+        lastReportedAvailability = currentTime
       }
-    case ReceiveTimeout =>
-      context.system.eventStream.publish(Unavailable(sourceEndpointId, logName))
+      schedule.cancel()
+      schedule = scheduleFailureDetectionLimitReached()
+      causes = Vector.empty
+    case FailureDetected(cause) =>
+      causes = causes :+ cause
```
volkerstampa: Alternatively we could publish `Unavailable` if the timespan between the first and the last failure in this list exceeds `failureDetectionLimit` (given that the list is emptied on each `AvailabilityDetected`). We would not need any scheduler then, right?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
krasserm: `System.nanoTime` is the recommended approach to measure elapsed time.

krasserm:
@volkerstampa with the last commit, I limited exception serialization to subclasses of `ReplicationReadException`.

PR #254, opened 2016-04-18
Local replication filters

### LINKED ISSUE #245: Support for local replication filters
In addition to `ReplicationConnection`-scoped replication filters, allow applications to define local `ReplicationEndpoint`-scoped replication filters. This allows replication sources to define which events can be replicated at all. Local replication filters should be configurable per event log name.

For example, a replication endpoint that doesn't want local events of type `X` to be replicated to other locations should setup a local replication filter that doesn't accept events of type `X`. This frees other replication endpoints to specify these filters in their replication connections. A main use case for local replication filters is to separate location-specific events from replicated events.

krasserm:
- closes #245


volkerstampa:
LGTM

PR #256, opened 2016-04-25
Observed-remove shopping cart CRDT

### LINKED ISSUE #114: Implement an OR Shopping Cart CRDT
Implement an OR-Map as defined in Spec 21 in the paper: [A comprehensive study of Convergent and Commutative Replicated Data Types](https://hal.inria.fr/inria-00555588/document).

An OR-Map is an extension of an OR-Set, where each item in the set is a key value mapping. However, unlike a set the key value mappings must be distinct. The OR-Map in Spec 21 defines the method to handle concurrent adding and removal of key value mappings, but it must also be possible to merge values when they are modified concurrently. In the case of Spec 21, an integer is chosen as the value type and two concurrent add's are merged by summing the integer of each add - so the value type is a monotonic counter. Since the values of an OR-Map must CRDTs themselves, it is necessary to implement a simple register to hold values first. As it happens, we already have an Multi Value Register (MVRegister) and hopefully a Last Write Wins Register (LWWRegister).

Some refactoring may be necessary to enable CmRDTs to be nested within each other, such that two concurrently modified registers can be merged.
An additional discussion on CRDTs is presented by Marc Shapiro in his [presentation at Microsoft Research](http://research.microsoft.com/apps/video/default.aspx?id=153540&r=1)

krasserm:
- closes #114

### DIFF
--- a/eventuate-crdt/src/it/scala/com/rbmhtechnology/eventuate/crdt/CRDTSerializerSpec.scala
+++ b/eventuate-crdt/src/it/scala/com/rbmhtechnology/eventuate/crdt/CRDTSerializerSpec.scala
@@ -27,6 +27,12 @@ object CRDTSerializerSpec {
   def orSet(payload: ExamplePayload) =
     ORSet[ExamplePayload].add(payload, VectorTime("s" -> 17L))
 
+  def orCart(key: ExamplePayload) =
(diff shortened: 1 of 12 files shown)


volkerstampa:
LGTM

Issue #142, opened 2015-11-21
Dynamic replication networks


krasserm:
It should be possible to add/remove replication connections to/from replication endpoints at runtime.

krasserm:
This ticket also includes the functionality to dynamically turn replication per connection on and off, so that uni-directional replication during application upgrade (#107) can be supported.

krasserm:
Covered by #314. Might become a ticket later but closing for now until decomposition of #314 into work packages becomes more stable.

coding4m:
@krasserm is there any plan to release this feature ?

krasserm:
@coding4m most likely with 0.9, beginning of 2017

PR #261, opened 2016-05-13
Failed disaster recovery should indicate whether partial updates have been made

### LINKED ISSUE #260: Failed disaster recovery should indicate whether partial updates have been made

krasserm:
- closes #260

### DIFF
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
@@ -267,10 +267,13 @@ class ReplicationEndpoint(
    * have been lost.
    *
    * This procedure requires that event replication between this and directly connected endpoints is bi-directional
-   * and that these endpoints are available during recovery. Recovery is idempotent and must be repeated in failure
-   * cases by the application. After successful recovery the endpoint is automatically activated.
-   * Activating this endpoint without having successfully recovered from partial or total event
-   * loss may result in inconsistent replica states.
+   * and that these endpoints are available during recovery. After successful recovery the endpoint is automatically
+   * activated. A failed recovery completes with a [[RecoveryException]] and must be retried. Activating this endpoint
+   * without having successfully recovered from partial or total event loss may result in inconsistent replica states.
+   *
+   * Running a recovery on an endpoint that didn't loose events has no effect but may still fail due to unavailable
+   * replication partners, for example. In this case, a recovery retry can be omitted if the `partialUpdate` field
+   * of [[RecoveryException]] is set to `false`.
    */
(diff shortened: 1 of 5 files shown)


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -281,11 +284,15 @@ class ReplicationEndpoint(
...
+      }
+
       val recoveryOperation = for {
-        infos <- recovery.readEndpointInfos
-        clocks <- recovery.readEventLogClocks
+        infos <- recovery.readEndpointInfos.recoverWith(recoveryFailure(partialUpdate = false))
+        clocks <- recovery.readEventLogClocks.recoverWith(recoveryFailure(partialUpdate = false))
         links = recovery.recoveryLinks(infos, clocks)
-        _ <- recovery.deleteSnapshots(links)
+        _ <- recovery.deleteSnapshots(links).recoverWith(recoveryFailure(partialUpdate = true))
```
magro: Just to be sure: Is `partialUpdate = true` here because the application might have deleted events under the assumption that the state is stored in the snapshot?

magro:
What about a minor doc update?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
krasserm: Snapshots that cover events that have been lost during a disaster must be removed. This is what  `recovery.deleteSnapshots(links)` is doing. If this operation fails, `partialUpdate` is set to `true` because it might have partially deleted some local snapshots already.

krasserm:
I'm a bit hesitant adding it to the reference documentation, as I'm not sure yet how disaster recovery will evolve in the next release (discussion to be done). It's anyway documented in the API docs. I'd rather like to extend the reference documentation later when we agreed upon how to proceed with disaster recovery in context of #142.

volkerstampa:
LGTM

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
magro: Ok. Learning eventuate more and more.

Btw: would it be a possible optimization to compare received `infos` and `clocks` with the locally stored data and abort/skip recovery (and this snapshot deletion)?

krasserm: Yes, we could consider that when evolving disaster recovery.

krasserm:
Thanks for the review @magro and @volkerstampa

PR #263, opened 2016-05-18
EventsourcedView and EventLog define their own logging adapter

### LINKED ISSUE #259: EventsourcedView should not extend ActorLogging
`EventsourcedView extends ActorLogging`, which currently makes it hard to create actors extending `DiagnosticActorLogging` - which is required to be able to set MDC values.

Those actors then require an additional `override val log: DiagnosticLoggingAdapter = akka.event.Logging(this)` in order to solve the conflict caused by inheriting both `ActorLogging.log` and `DiagnosticActorLogging.log`.

One simple solution would be to let `EventsourcedView` extend `DiagnosticActorLogging` instead, because the provided `val log: DiagnosticLoggingAdapter` extends the currently used `LoggingAdapter`.

But there are also other `LoggingAdapter` implementations (`BusLogging`, `NoLogging`), so it might make sense to make this somehow configurable?

@krasserm, what do you think?

krasserm:
- avoid name conflicts with logging adapter of ActorLogging and related traits
- closes #259

### DIFF
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedView.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedView.scala
@@ -20,6 +20,7 @@ import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicInteger
 
 import akka.actor._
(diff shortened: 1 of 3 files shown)


volkerstampa:
LGTM

PR #276, opened 2016-06-23
verify gap-less sequence numbers in leveldb eventlog

### LINKED ISSUE #274: Verify gap-less stored sequence in LevelDB event log
`LeveldbEventLog` is designed to generate a gap-less local sequence. This should be verified on replication read. From a recent failure scenario in a project, we suspected a failed write that wasn't reported to the application (= event log) even with `fsync = on` which would generate a stored sequence with gaps.

kongo2002:
closes #274


krasserm:
LGTM 😃

PR #278, opened 2016-06-27
Stateful processor calls processEvent for previously processed events

### LINKED ISSUE #277: StatefulProcessor doesn't properly recover state

morsch:
This works with little to no overheard because regular (ie. non-stateful) processors won't be called for previously processed events. Stateful processors return None in `readSuccess`, so `onEvent` will be called for old events (snapshots aside). That said, the interaction between the var `processingProgress` and the return value of `readSuccess` is a bit tricky.


krasserm:
LGTM, good catch! Thanks for the fix @morsch

PR #284, opened 2016-07-04
Upgrade to Cassandra 3.4 for testing

### LINKED ISSUE #149: Upgrade to Cassandra 3.x

krasserm:
- closes #149


volkerstampa:
I wonder if it is worth to introduce a new config for multi-jvm tests (for the cassandra sub-project) that extends the IntergrationTest config to get rid of the redundant EmbeddedCassandra.scala/log4j2.xml/test-cassandra.yaml. Could maybe also be handled in a separate ticket?!?

Other than that LGTM.

krasserm:
Thanks for reviewing @volkerstampa. I tried that extension before but this somehow messed up the classpath in IDEA, so I decided to go for that (small) redundancy to work around that problem. I couldn't find the root cause but I'll create a separate ticket to further investigate.

PR #294, opened 2016-07-27
Support locally defined filter per target log id

### LINKED ISSUE #273: Replication connection specific local replication filters
#245 introduced endpoint-global replication filters that apply to all inbound connections. It should be possible to define local replication filters per inbound connection, using a default replication filter if no connection specific filter is defined. It should also be investigated if connection specific local replication filters can make remote replication filters obsolete (which is desired).

volkerstampa:
Locally defined replication filters can be defined by log name and by
target log id. Those defined for target log id take precedence.

Closes #273


magro:
LGTM. So remote replication filters shall not be made  obsolete, right?

volkerstampa:
@magro Thanks for the review. I thought it makes sense to keep remote replication filters at least for the moment to avoid immediate incompatibilities with existing software.

magro:
Sounds good, thanks!

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -216,8 +216,9 @@ object ReplicationEndpoint {
  * @param logFactory Factory of log actor `Props`. The `String` parameter of the factory is a unique
  *                   log id generated by this endpoint. The log actor must be assigned this log id.
  * @param connections Replication connections to other replication endpoints.
- * @param filters Replication filters applied locally. Filters are applied to individual
- *                event logs where filter keys are the corresponding event log names.
+ * @param filters Replication filters applied locally. Filters are applied when replicating to
+ *               individual target logs (keys are log ids of remote logs) or from individual
+ *               local logs to all corresponding targets logs (keys are log-names).
```
krasserm: I think we should have two filter maps here: one for maintaining filters per (local/source) log name (`sourceFilters`) and another one for maintaining filter per (remote/target) log id (`targetFilters`). This approach doesn't mix map keys with different semantics in a single map.

In addition to replacing a source filter with a target filter, if defined, we should also support use cases where a source filter is AND-combined with a target filter. An example is to have a source filter that separates "private" events (that should never get replicated) from public events (that can be replicated to other locations).

PR #295, opened 2016-07-28
Add info about stable destination ActorPaths

volkerstampa:


magro:
It could additionally mentioned that the actor must actually exist, therfore it must not be a lazy val. Perhaps an example in the line of "instead of this do this" would be helpful?

magro:
I just checked the current [documentation on reliable delivery](http://rbmhtechnology.github.io/eventuate/reference/event-sourcing.html#reliable-delivery): it uses a `destination: ActorPath` and does not show the side that passes this `ActorPath` to the `ExampleActor`. Then it would make sense to add this side to the example as well, so that the comment regarding the lazy actor does not cause confusion.

volkerstampa:
@magro In the documentation `ConfirmedDelivery` is mentioned as a tool in context of remote communication, so you can never be sure if the destination actor actually exists (or is able to receive the message). That is why you should invoke `redeliverUnconfirmed` on a regular basis (as shown in the example).

magro:
Right, makes sense. So I also don't see the need for further adjustments.

krasserm:
LGTM

PR #298, opened 2016-08-01
Change intermediate replication failure logs from error to warning

magro:
...and add an ERROR log for the final failure when the failure detection limit is reached.

**Motivation**

We're regularly observing "replication read failed" ERRORs logged by
`com.rbmhtechnology.eventuate.Replicator`, caused by a

    com.rbmhtechnology.eventuate.ReplicationProtocol$ReplicationReadTimeoutException: Replication read timed out after 30000 milliseconds

While of cause this highlights an issue with the underlying network,
it's nothing special because the network is just not reliable.
These cases are also gracefully handled by eventuate because it retries
and only when the failure detection limit is reached the replication endpoint
is considered and reported as unavailable.

Logs with level ERROR would also require our ops team to take some action
or at least it would need some attention, while we'd say this is not necessary.

Therefore the log level for temporary failures is changed from ERROR
to WARN. Additionally the exception's stack trace is omitted, because it's
not carrying much useful information, the exception + message should be sufficient
(additionally akka's logger simply doesn't allow to provide an exception for
a warning).  For reference, the stack trace is s.th. like this:

    com.rbmhtechnology.eventuate.ReplicationProtocol$ReplicationReadTimeoutException: Replication read timed out after 30000 milliseconds
            at com.rbmhtechnology.eventuate.Replicator$$anonfun$com$rbmhtechnology$eventuate$Replicator$$read$1.applyOrElse(ReplicationEndpoint.scala:543) ~[com.rbmhtechnology.eventuate-core_2.11-0.7.1.jar:0.7.1]
            at com.rbmhtechnology.eventuate.Replicator$$anonfun$com$rbmhtechnology$eventuate$Replicator$$read$1.applyOrElse(ReplicationEndpoint.scala:542) ~[com.rbmhtechnology.eventuate-core_2.11-0.7.1.jar:0.7.1]
            at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36) ~[org.scala-lang.scala-library-2.11.8.jar:na]
            at scala.util.Failure$$anonfun$recover$1.apply(Try.scala:216) ~[org.scala-lang.scala-library-2.11.8.jar:na]
            at scala.util.Try$.apply(Try.scala:192) ~[org.scala-lang.scala-library-2.11.8.jar:na]
            at scala.util.Failure.recover(Try.scala:216) ~[org.scala-lang.scala-library-2.11.8.jar:na]
            at scala.concurrent.Future$$anonfun$recover$1.apply(Future.scala:326) ~[org.scala-lang.scala-library-2.11.8.jar:na]
            at scala.concurrent.Future$$anonfun$recover$1.apply(Future.scala:326) ~[org.scala-lang.scala-library-2.11.8.jar:na]
            at scala.concurrent.impl.CallbackRunnable.run_aroundBody0(Promise.scala:32) ~[org.scala-lang.scala-library-2.11.8.jar:na]
            at scala.concurrent.impl.CallbackRunnable$AjcClosure1.run(Promise.scala:1) ~[org.scala-lang.scala-library-2.11.8.jar:na]

The final failure then is considered being an issue which should in fact
get some attention, therefore an additional ERROR log is added for this case.
The last exception is also provided for this ERROR log, or akka's `Error.NoCause`
if no last exception available (akka handles `Error.NoCause` in a way as if
`error(message)` was called).

### DIFF
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
@@ -21,6 +21,7 @@ import java.util.concurrent.atomic.AtomicBoolean
 import java.util.function.{ Function => JFunction }
 
 import akka.actor._
+import akka.event.Logging
 import akka.pattern.ask
 import akka.pattern.pipe
 import akka.util.Timeout
@@ -474,7 +475,7 @@ private class Replicator(target: ReplicationTarget, source: ReplicationSource, f
       context.become(reading)
       read(storedReplicationProgress, currentTargetVersionVector)
     case GetReplicationProgressFailure(cause) =>
-      log.error(cause, "replication progress read failed")
+      log.warning("replication progress read failed: {}", cause)
       scheduleFetch()
   }
 
@@ -492,7 +493,7 @@ private class Replicator(target: ReplicationTarget, source: ReplicationSource, f
       write(events, replicationProgress, currentSourceVersionVector, replicationProgress >= fromSequenceNr)
     case ReplicationReadFailure(cause, _) =>
       detector ! FailureDetected(cause)
-      log.error(cause, s"replication read failed")
+      log.warning(s"replication read failed: {}", cause)
       context.become(idle)
       scheduleRead()
   }
@@ -507,7 +508,7 @@ private class Replicator(target: ReplicationTarget, source: ReplicationSource, f
       context.become(reading)
       read(storedReplicationProgress, currentTargetVersionVector)
     case ReplicationWriteFailure(cause) =>
-      log.error(cause, "replication write failed")
+      log.warning("replication write failed: {}", cause)
       context.become(idle)
       scheduleRead()
   }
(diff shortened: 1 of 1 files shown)


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -492,7 +493,7 @@ private class Replicator(target: ReplicationTarget, source: ReplicationSource, f
       write(events, replicationProgress, currentSourceVersionVector, replicationProgress >= fromSequenceNr)
     case ReplicationReadFailure(cause, _) =>
       detector ! FailureDetected(cause)
-      log.error(cause, s"replication read failed")
+      log.warning(s"replication read failed: {}", cause)
```
volkerstampa: The `s` was there before, but it is actually not needed, right?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -592,6 +593,8 @@ private class FailureDetector(sourceEndpointId: String, logName: String, failure
     case FailureDetected(cause) =>
       causes = causes :+ cause
     case FailureDetectionLimitReached(scheduledCount) if scheduledCount == counter =>
+      log.error(causes.lastOption.getOrElse(Logging.Error.NoCause), "replication failure detection limit reached ({})," +
+        " publishing Unavailable for {}/{} (last exception being reported)", failureDetectionLimit, sourceEndpointId, logName)
```
volkerstampa: another nitpick: avoid splitting the string with `+` and start the new line before `"replication...`

volkerstampa:
LGTM

krasserm:
LGTM

PR #300, opened 2016-08-23
Replay from application-defined sequence number

danbim:
We have a use-case in which we want to speed up recovery and minimize snapshot size (currently it is around 700mb, its size being determined by the days of transaction history being kept in memory).

To minimize the snapshot size and reduce the replay duration the actor memorizes before which point in time (= log sequence number) historical data "can be forgotten". The log sequence number is therefore persisted as (part of) its snapshot. The "forgotten data" does therefore not need to be replayed during recovery, reducing replay/recovery (down)time.

### DIFF
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedWriter.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedWriter.scala
@@ -89,6 +89,12 @@ trait EventsourcedWriter[R, W] extends EventsourcedView {
   private var numPending: Int = 0
 
   /**
+   * Disallow for [[EventsourcedWriter]] and subclasses as event processing progress is determined by `read`
+   * and `readSuccess`.
+   */
+  override final def replayFromSequenceNr: Option[Long] = None
+
+  /**
    * Asynchronously reads an initial value from the target database, usually to obtain information about
    * event processing progress. This method is called during initialization.
    */
@@ -115,7 +121,7 @@ trait EventsourcedWriter[R, W] extends EventsourcedView {
    * and can be overridden.
    */
   def readSuccess(result: R): Option[Long] =
-    None
+    replayFromSequenceNr
 
   /**
    * Called with a write result after a `write` operation successfully completes. This method may update
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedView.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedView.scala
@@ -308,7 +308,7 @@ trait EventsourcedView extends Actor with Stash {
     val iid = instanceId
 
     if (saveRequests.contains(metadata)) {
-      handler(Failure(new IllegalStateException(s"snapshot with metadata ${metadata} is currently being saved")))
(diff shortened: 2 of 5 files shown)


krasserm:
Thanks for the pull request @danbim. Replaying from a custom position is definitely a missing feature in Eventuate and should be added. However, I think it should not be coupled to snapshot loading for two reasons:

- it doesn't allow replay from a custom position when there's no snapshot for that actor
- it may break consistent recovery of actor state when the replay position is not `snapshot.metadata.sequenceNr + 1L`

I want to propose the following alternative implementation for `EventsourcedView` (which is what `EventsourcedWriter` already implements):

- read a custom start sequence number from somewhere. This can be a custom data source, a constructor argument or whatever.
- if a custom start sequence number could be read (`Some(snr)`), start event replay from `snr`, otherwise (`None`) start the default snapshot loading procedure. If no snapshot exists, start replay from sequence number `1L`, otherwise start replay from `snapshot.metadata.sequenceNr + 1L`.

This would solve the issues described above and allows to implement `EventsourcedWriter` in terms of `EventsourcedView`. If you additionally need to initialize actor state with a custom value, you can do that via the actor's constructor.

WDYT?

danbim:
Sounds pretty good to me 👍 AFAICS. I'll try to update the PR soon

danbim:
@krasserm I've been trying to implement what you suggested (see updated PR) but I'm not sure how to implement ```EventsourcedWriter``` in terms of ```EventsourcedView```. ```EventsourcedWriter``` is type-parameterized for the type of the read operation while ```EventsourcedView``` is not.

I'm gonna do some thinking / reading on this. Meanwhile, maybe you have a hint for me? :)

danbim:
Had an attempt at implementing ```EventsourcedWriter``` in terms of ```EventsourcedView```. However, the implementation doesn't really feel "elegant" to me yet. WDYT?

krasserm:
A thinking a while, I propose to follow [this approach](https://github.com/RBMHTechnology/eventuate/commit/1703405eef3f6506f5a9c193b8dad2c3ad006f67). It's much easier than my initial implementation proposal, still has all its the advantages. Not documented/tested yet but I hope the idea is clear. Thoughts?

danbim:
Looks similar to mine with the one difference: in your proposal the "calculation" of the replaySequenceNr is a synchronous process.

My assumption was that typically an EventsourcedView would have to retrieve the replaySequenceNr from some kind of storage or other (remote) service and would therefore require to do so asynchronously, thus the Future return type (which might also fail, obviously).

Do you think my assumption is wrong?

krasserm:
Please see the comments in https://github.com/RBMHTechnology/eventuate/issues/302

danbim:
Aah! I didn't see that discussion yet. Let's continue over there...

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedWriter.scala
```diff
@@ -88,6 +88,8 @@ trait EventsourcedWriter[R, W] extends EventsourcedView {
 
   private var numPending: Int = 0
 
+  override final def replayFromSequenceNr: Option[Long] = None
```
krasserm: Please add an API doc that explains what is returned and why.

krasserm:
LGTM, thanks a lot! I'll merge after the API doc has been added.

Issue #297, opened 2016-07-29
Deprecate remote replication filters


volkerstampa:
#273 just added connection specific local replication filters. If remote replication filters prove to be obsolete support for them should be removed.

PR #303, opened 2016-09-06
Allow advanced combination of local replication filters

### LINKED ISSUE #273: Replication connection specific local replication filters

volkerstampa:
Local replication filters can be specified per source log name or
target log id in form of an EndpointFilter.
There are EndpointFilters for source or target logs only or
logical combination of them.

- closes #273


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -363,6 +359,96 @@ class ReplicationEndpoint(
...
+  /**
+   * Creates an [[EndpointFilter]] that computes a [[ReplicationFilter]] for a replication read request
+   * from a source log to a target log by returning the target filter when given in `targetFilters`. If only
+   * a source filter is given in `sourceFilters` that is returned otherwise [[com.rbmhtechnology.eventuate.ReplicationFilter.NoFilter]] is returned.
+   *
+   * This [[EndpointFilter]] can be useful when an eventsourced service that can be replicated collaborates
+   * with different internal services and has a fixed set of events for external communication.
+   * In that case `sourceFilters` can ensure that only external events are replicated to external services
+   * while in case of internal communication (with internal service or replication partners) these filters
+   * can be overwritten.
```
krasserm: I think most people will have difficulties to understand what you mean with *internal* and *external* here. This comes from an application domain but it is not a concept of Eventuate itself. I'd rather say that the use case is that some (more privileged) remote targets may replace a (default) source filter with a target-specific filter. In this context, I also find the term *service* a bit confusing. To me it would be clearer if the API docs mention *local source (log)* and *remote target (log)*.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -363,6 +359,96 @@ class ReplicationEndpoint(
...
+   *
+   * @param targetFilters maps target log ids to the [[ReplicationFilter]] that shall be applied when replicating to the target log
+   * @param sourceFilters maps source log names to the [[ReplicationFilter]] that shall be applied when replicating from the source log
```
krasserm: Distinguishing *replicating to the target log* from *replicating from the source log* is a bit misleading here. In both case there is a replication from the source log to the target log.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -363,6 +359,96 @@ class ReplicationEndpoint(
...
+  }
+
+  /**
+   * Creates an [[EndpointFilter]] that computes a [[ReplicationFilter]] for a replication read request
+   * from a source log to a target log by and-combining target and source filters when given in the provided [[Map]]s.
+   * If only source or target filter is given that is returned. If no filter is given [[com.rbmhtechnology.eventuate.ReplicationFilter.NoFilter]] is returned.
+   *
+   * This [[EndpointFilter]] can be useful when an eventsourced service that is not replicated collaborates
+   * with different external service. In that case `sourceFilters` can ensure that no ''internal'' event is
+   * replicated to other services and `targetFilters` can add target specific filter criteria.
```
krasserm: See also https://github.com/RBMHTechnology/eventuate/pull/303/commits/05f58abd6c0d37d8fe44e97277989355e0a8fb80#r77624524 regarding the term *service*

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -363,6 +359,96 @@ class ReplicationEndpoint(
...
 /**
+ * An [[EndpointFilter]] computes a [[ReplicationFilter]] that shall be applied to a
+ * replication read request that replicates from a source log (defined by ``sourceLogName``)
+ * to a target log (defined by ``targetLogId``).
+ */
+trait EndpointFilter {
+  def filterFor(targetLogId: String, sourceLogName: String): ReplicationFilter
+}
+
+object EndpointFilter {
```
krasserm: To be more consistent with existing variable names (and documentation) `EndpointFilter` should be renamed to `EndpointFilters`. However, I would prefer the name `ReplicationFilterCombination` or `ReplicationFilterCombiner` or somthing similar. WDYT?

eventuate-core/src/it/scala/com/rbmhtechnology/eventuate/ReplicationIntegrationSpec.scala
```diff
@@ -162,8 +165,8 @@ trait ReplicationIntegrationSpec extends WordSpec with Matchers with MultiLocati
       val locationA = location("A")
       val locationB = location("B")
 
-      val endpointA = locationA.endpoint(Set("L1"), Set(replicationConnection(locationB.port, filters = Map("L1" -> new PayloadInequalityFilter("b1")))), filters = Map("L1" -> new PayloadInequalityFilter("a3")))
-      val endpointB = locationB.endpoint(Set("L1"), Set(replicationConnection(locationA.port, filters = Map("L1" -> new PayloadInequalityFilter("a1")))), filters = Map("L1" -> new PayloadInequalityFilter("b3")))
+      val endpointA = locationA.endpoint(Set("L1"), Set(replicationConnection(locationB.port, filters = Map("L1" -> new PayloadInequalityFilter("b1")))), sourceFilters(Map("L1" -> new PayloadInequalityFilter("a2"))))
+      val endpointB = locationB.endpoint(Set("L1"), Set(replicationConnection(locationA.port, filters = Map("L1" -> new PayloadInequalityFilter("a1")))), sourceFilters(Map("L1" -> new PayloadInequalityFilter("b2"))))
```
krasserm: Is there a special reason that `a3` and `b3` have been replaced by `a2` and `b2`?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
```diff
@@ -183,10 +182,7 @@ private class Acceptor(endpoint: ReplicationEndpoint) extends Actor {
     case re: ReplicationReadEnvelope if re.incompatibleWith(endpoint.applicationName, endpoint.applicationVersion) =>
       sender ! ReplicationReadFailure(IncompatibleApplicationVersionException(endpoint.id, endpoint.applicationVersion, re.targetApplicationVersion), re.payload.targetLogId)
     case ReplicationReadEnvelope(r, logName, _, _) =>
-      val r2 = endpoint.filters.get(r.targetLogId).orElse(endpoint.filters.get(logName)) match {
-        case Some(filter) => r.copy(filter = r.filter and filter)
-        case None         => r
-      }
+      val r2 = r.copy(filter = endpoint.endpointFilters.filterFor(r.targetLogId, logName) and r.filter)
```
krasserm: `and r.filter` can be removed once https://github.com/RBMHTechnology/eventuate/issues/297 is resolved, right?

krasserm:
LGTM, nice solution!

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
volkerstampa: You only talk about the `object` not the `trait`, right? To me `ReplicationFilterCombiner` (or similar) sounds like something that generically combines `ReplicationFilter`s however this is done with a `ReplicationFilter`'s methods (like and, or, ...). `EndpointFilter` is basically a factory for `EndpointFilter`s, so I would prefer `EndpointFilters`. OK?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
volkerstampa: Right.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
krasserm: I talked about both, the trait and companion object. Anyway, `EndpointFilters` is fine for me.

PR #304, opened 2016-09-07
implement per-eventlog cassandra settings

kongo2002:
Hi,

at first I  planned to implement a configurable per-eventlog `partition-size` only in order to cope with heterogenous event sizes across different event logs. Now you can override the  cassandra settings as a whole.

The `CassandraEventLog` constructor and its companion's `props` method
support an optional argument of type `Option[CassandraEventLogSettings]`
to override the 'global' settings of the `Cassandra` extension for a
specific eventlog.

As soon as this is worth integrating we can mention this 'feature' in the appropriate sections in the docs.

Cheers,
Gregor

### DIFF
--- a/eventuate-adapter-spark/src/it/scala/com/rbmhtechnology/eventuate/adapter/spark/SparkBatchAdapterSpec.scala
+++ b/eventuate-adapter-spark/src/it/scala/com/rbmhtechnology/eventuate/adapter/spark/SparkBatchAdapterSpec.scala
@@ -18,22 +18,20 @@ package com.rbmhtechnology.eventuate.adapter.spark
 
 import akka.actor._
 import akka.testkit._
-
 import com.rbmhtechnology.eventuate._
 import com.rbmhtechnology.eventuate.utilities._
 import com.rbmhtechnology.eventuate.log.EventLogWriter
-import com.rbmhtechnology.eventuate.log.cassandra.CassandraEventLogSettings
-
+import com.rbmhtechnology.eventuate.log.cassandra.CassandraSettings
 import org.apache.spark._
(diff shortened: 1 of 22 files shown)


krasserm:
Overriding certain config parameters per event log is definitely a valid use case but I'm not happy with the semantics of the proposed solution. Having an optional `CassandraEventLogSettings` per event log, a user might expect that he can also override the `keyspace` or `connect-retry-max` configuration settings, for example, but this is actually not the case. Hence, a 👎 on your proposal.

An implementation with clear semantics should distinguish global settings from log-specific settings in the `*Settings` type. However, this would mean a bigger redesign. Alternatively, you could allow event log instances to override the `partition-size`. For example, this could be done by creating a subclass of `CassandraEventLog` that overrides a default `partitionSize` method. Alternatively, the partition size could be a constructor parameter. 

WDYT?

kongo2002:
Ok, I get your point. Originally I just wanted to add a `partitionSize` constructor parameter only but wasn't sure what this would lead to when you want to add more eventlog specific configuration parameters. That's why I went for the whole settings approach this time.

kongo2002:
@krasserm I've just pushed the discussed alternative approach of splitting the Cassandra related settings into global (`CassandraSettings`) and event log specific ones (`CassandraEventLogSettings`). 

Just let me know if I am getting closer :D

krasserm:
@kongo2002 any plans to work on that ticket further?

kongo2002:
@krasserm Sorry for the delayed response. Right now I don't see me having spare time to push this PR any further. Although I think this is a reasonable feature to integrate, for now I do have a work-around for this 'issue' in-place.

Issue #183, opened 2015-12-31
Investigate logged NPE during normal test run


krasserm:
First seen at https://travis-ci.org/RBMHTechnology/eventuate/builds/99603209#L1649

krasserm:
Seen again at https://travis-ci.org/RBMHTechnology/eventuate/builds/100780601#L1635

PR #308, opened 2016-09-12
Wip 286 disaster recovery enhancements

volkerstampa:


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -509,11 +532,13 @@ private class Connector(sourceConnector: SourceConnector) extends Actor {
     case GetReplicationEndpointInfoSuccess(info) if !connected =>
       sourceConnector.links(info).foreach {
         case ReplicationLink(source, target) =>
-          val filter = sourceConnector.connection.filters.get(target.logName) match {
-            case Some(f) => f
-            case None    => NoFilter
+          if (sourceLogIds.forall(_.contains(source.logId))) {
```
krasserm: Why not having a `replicationLinks: Set[ReplicationLink]` parameter instead if callers want to control which replicators should be set up? All callers have access to the `SourceConnector.links` method to obtain the links they need.

eventuate-core/src/main/protobuf/ReplicationProtocolFormats.proto
```diff
@@ -44,6 +44,22 @@ message GetReplicationEndpointInfoSuccessFormat {
...
+message SynchronizeReplicationProgressSuccessFormat {
+  optional ReplicationEndpointInfoFormat info = 1;
+}
+
+message SynchronizeReplicationProgressFailureFormat {
+  optional PayloadFormat cause = 1;
+}
+
+message SynchronizeReplicationProgressSourceExceptionFormat {
+  optional string causeMessage = 1;
```
krasserm: Just `message`, as used by most other exception classes?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
```diff
@@ -152,30 +199,26 @@ private class Acceptor(endpoint: ReplicationEndpoint) extends Actor {
...
+      context.become(processing)
   }
 
   def recoveringEvents(recoveryManager: ActorRef, promise: Promise[Unit]): Receive = {
     case writeSuccess: ReplicationWriteSuccess =>
       recoveryManager forward writeSuccess
     case EventRecoveryCompleted =>
       promise.success(())
-      context.become(processing)
+      context.become(recovering orElse processing)
```
krasserm: Why is `recovering orElse` necessary? Recovery is completed here.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -248,8 +248,10 @@ class ReplicationEndpoint(
   private[eventuate] val connectors: Set[SourceConnector] =
     connections.map(new SourceConnector(this, _))
 
+  // lazy to make sure concurrently running (created actors) do not access null-reference
   private[eventuate] lazy val acceptor: ActorRef =
     system.actorOf(Props(new Acceptor(this)), name = Acceptor.Name)
+  acceptor // make sure acceptor is started
```
krasserm: Why not making `acceptor` an "eager" `val` instead? It's anyway evaluated immediately here.

krasserm: I think we already discussed that in context of a bug but I don't remember the details any more. Do you?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationProtocol.scala
```diff
@@ -80,6 +67,38 @@ object ReplicationProtocol {
...
+
+  /**
+   * Base class for all [[SynchronizeReplicationProgressFailure]] `cause`s.
+   */
+  private[eventuate] abstract class SynchronizeReplicationProgressException(message: String) extends RuntimeException(message)
+
+  /**
+   * Indicates a problem synchronizing the replication progress of a remote [[ReplicationEndpoint]]
+   */
+  private[eventuate] case class SynchronizeReplicationProgressSourceException(causeMessage: String)
```
krasserm: Just `message`?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
```diff
@@ -279,28 +281,48 @@ class ReplicationEndpoint(
...
-      } yield acceptor ! Recover(links, promise)
+      for {
+        localEndpointInfo <- recovery.readEndpointInfo.recoverWith(recoveryFailure(partialUpdate = false))
+        _ = logLocalState(localEndpointInfo)
+        remoteEndpointInfos <- recovery.synchronizeReplicationProgressesWithRemote(localEndpointInfo).recoverWith(recoveryFailure(partialUpdate = false))
+        _ = logRemoteState(remoteEndpointInfos)
+        unfilteredLinks = recovery.recoveryLinks(remoteEndpointInfos, localEndpointInfo, filtered = false)
+        _ = logLinksToBeRecovered(unfilteredLinks, "unfiltered")
+        _ <- recovery.recoverLinks(unfilteredLinks).recoverWith(recoveryFailure(partialUpdate = true))
+        filteredLinks = recovery.recoveryLinks(remoteEndpointInfos, localEndpointInfo, filtered = true)
```
krasserm: It would be useful to document why unfiltered connections are used for recover first, then filtered connections.

krasserm: Also, the API docs should mention that after having successfully run `recover()` the endpoint is already activated i.e. applications don't need to run `activate()` afterwards.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/log/EventLog.scala
```diff
@@ -575,7 +575,7 @@ abstract class EventLog[A <: EventLogState](id: String) extends Actor with Event
         // 100% filtering coverage for certain replication network topologies.
         acc
       case (acc, e) =>
-        snr += 1L
+        snr = (snr + 1) max e.vectorTimestamp.value.getOrElse(id, 0)
```
krasserm: This breaks the algorithm for event log partitioning (used in `CassandraEventLog`). I propose to adjust the local sequence number from the log's version vector, if needed, after successful completion of recovery. This requires a new event log command for updating the local sequence number. The handler of this command should respect partition sizes (see also `EventLog.adjustSequenceNr`).

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
```diff
@@ -81,57 +83,102 @@ private class Recovery(endpoint: ReplicationEndpoint) {
...
 
-  private def readClocksImpl: Future[Map[String, EventLogClock]] = {
-    Future.traverse(endpoint.logNames)(name => readEventLogClock(endpoint.logs(name)).map(name -> _)).map(_.toMap)
+  private def readReplicationProgress(logActor: ActorRef, logId: String): Future[Long] =
+    readResult[GetReplicationProgressSuccess, GetReplicationProgressFailure, Long](
+      logActor.ask(GetReplicationProgress(logId)), _.storedReplicationProgress, _.cause)
+
+  private def updateReplicationProgress(logActor: ActorRef, logId: String, sequenceNr: Long): Future[Long] = {
+    readResult[ReplicationWriteSuccess, ReplicationWriteFailure, Long](
+      logActor.ask(ReplicationWrite(Seq.empty, sequenceNr, logId, VectorTime.Zero)), _.storedReplicationProgress, _.cause)
```
krasserm: Try using `ReplicationProtocol.SetReplicationProgress` here.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
volkerstampa: But they do not have access to the `ReplicationEndpointInfo` that they have to provide to the `links` methods, right? In particular `ReplicationEndpoint.activate` does not have access to an `ReplicationEndpointInfo`, but even in `Acceptor.recovering` the `links` might be across several endpoints. Or am I missing something?

eventuate-core/src/main/protobuf/ReplicationProtocolFormats.proto
volkerstampa: ok.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
volkerstampa: Just for the `links` provided in the `Recover` message, but there might be several `Recover` _invocations_ (one for filtered, one for unfiltered connections). Recovery is complete when `RecoveryFinished` is received.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
volkerstampa: The related issue is #183.  `this` is passed to `Acceptor` before construction of the `ReplicationEndpoint` completes and `Acceptor` is created and may reference `ReplicationEndpoint`s member `acceptor` concurrently (i.e. before it is initialized).

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationProtocol.scala
volkerstampa: OK.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
volkerstampa: The API doc actually states: `After successful recovery the endpoint is automatically activated.` I will add a comment why unfiltered connections have to be recovered first.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
volkerstampa: We also need to reset `EventLog.replicaVersionVectors`. I propose to rename this method to `updateReplicationMetadata` and add a comment. WDYT?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/log/EventLog.scala
volkerstampa: Good catch. Thanks for this. Will try your proposal.

volkerstampa: I actually tried something different. I moved the sequence number adjustment down to [line 585](https://github.com/RBMHTechnology/eventuate/blob/wip-286-disaster-recovery-enhancements/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/log/EventLog.scala#L585):

        (updated, clock.copy(sequenceNr = snr max lvv.value.getOrElse(id, 0), versionVector = lvv))

Not sure if it is better than your proposal, but as far as I understand it should not suffer from the problem you described and it is certainly much simpler. @kongo2002 In any case this all results in potential gaps in all event log implementations including the leveldb one, but currently we have a _gap-detecion_ algorithm in there that makes all event iterations with gaps fail. It was introduced with #274. I wonder if we should simply drop this gap detection or have to find something smarter. WDYT?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
krasserm: I see, ok. But setting it to `VectorTime.Zero` makes only sense if the remote location lost all events. If it has been restored from a backup, we would potentially transfer a large number of events just for being ignored at the target.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
krasserm: I see. Maybe rename `RecoveredFinished` to `RecoveryCompleted` to be consistent with `EventRecoveryCompleted`.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
krasserm: Ok. Please add a comment with a reference to a issue.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/log/EventLog.scala
krasserm: I think these adjustments should be made at the event of event recovery i.e. should be part of the disaster recovery code. It is not needed during normal event replication.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/Recovery.scala
volkerstampa: Yep. Makes sense.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/ReplicationEndpoint.scala
volkerstampa: Done

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/log/EventLog.scala
volkerstampa: Changes the implementation accordingly. See the newest commit on the PR.

krasserm:
LGTM

PR #310, opened 2016-09-13
Vert.x Adapter

### LINKED ISSUE #290: Vert.x Adapter
Create an adapter to integrate Eventuate with [Vert.x](http://vertx.io/). The primary interface for the integration is the Vert.x [event bus](http://vertx.io/docs/vertx-core/java/#event_bus).

The adapter should support following use-cases:
- publish events from an event log to multiple consumers on the Vert.x event bus
- send events from an event log to a single consumer with reliable delivery guarantees
- persist events from the Vert.x event bus to an event log

cstub:
The Vert.x Adapter offers various facilities to connect Eventuate event-logs to the Vert.x event-bus. Events can be read from an event-log and published to the Vert.x event-bus. Furthermore events may be read from the Vert.x event-bus and persisted to an event-log.

- implement an adapter-system which is provided with adapter-configurations and an instance of Vert.x
- implement a publish-adapter which reads events from an event-log and publishes them to all consumers on the Vert.x event-bus
- implement a send-adapter which reads events from an event-log and sends them to a single consumer on the Vert.x event-bus
- provide at-least-once-delivery semantics for the send-adapter with single confirmations or batched confirmations
- implement a write-adapter which reads events from the Vert.x event-bus and persists them to an event-log

- closes #290


krasserm:
When compiling I'm getting lots of

```
[warn] Class io.vertx.codegen.annotations.VertxGen not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.GenIgnore not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.GenIgnore not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.GenIgnore not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.GenIgnore not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.Fluent not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.Fluent not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.GenIgnore not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.Fluent not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.GenIgnore not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.Fluent not found - continuing with a stub.
[warn] Class io.vertx.codegen.annotations.Fluent not found - continuing with a stub.
...
```

How shall we deal with it?

krasserm:
Please add this adapter to the [build config](https://github.com/RBMHTechnology/eventuate/blob/wip-290-vertx-adapter/.travis.yml).

project/ProjectDependencies.scala
```diff
@@ -22,6 +22,8 @@ object ProjectDependencyVersions {
   val Log4jVersion = "2.5"
   val ProtobufVersion = "2.5.0"
   val SparkVersion = "1.6.1"
+  val VertxVersion = "3.0.0"
+  val ExampleVertxVersion = "3.3.2"
```
krasserm: Why different versions for the adapter and the example?

cstub: I choose different versions because the Adapter is compatible with Vert.x versions > 3.0.0, but version 3.0.0 has some logging issues and because of that the output in the examples was messed up.

krasserm: Assuming that 3.3.2 is backwards-compatible to 3.0.0, why not depending on 3.3.2 in the adapter then?

cstub: It is backwards compatible, but using 3.3.2 in the adapter also has the risk of using APIs not available in Vert.x 3.0.0, which would break the compatibility of the adapter, wouldn't it?

cstub: I can also try using Vert.x 3.0.0 in the example and disable the logging.

krasserm:
`build.sbt` seems to break the diff view on github, hence my comments here:

- `dependsOn(logLeveldb % "it->it")` should be enough, the Cassandra dependencies can be removed
- The Netty dependency of Akka Remote must not be excluded as it depends on Netty 3 which cannot be replaced by Netty 4. I wonder if it is a problem at all to have both Netty 3 and 4 on the classpath, as Netty 3 is in the `org.jboss.netty` package and Netty 4 in the `io.netty` package.
- The multi-jvm configs/dependencies can be removed as well unless you plan to make multi-jvm tests (see also the [Akka Streams adapter configuration](https://github.com/RBMHTechnology/eventuate/blob/wip-98-akka-streams-adapter/build.sbt#L75) for a more lightweight example)

project/ProjectDependencies.scala
krasserm: Isn't that only an issue when using a `provided` configuration. If we force applications to use 3.3.2 this should be ok, unless there are breaking changes in Vert.x 3.x minor versions. Is this the case?

cstub: The adapter should use Vert.x 3.0.0 as a provided dependency to not force the user into a specific Vert.x version because of breaking changes between minor versions (see https://github.com/vert-x3/wiki/wiki/3.1.0---Breaking-changes).

krasserm: Ok, I see ...

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
```diff
@@ -0,0 +1,68 @@
...
+
+class VertxAdapterSystemConfig(private[vertx] val configurations: Seq[VertxAdapterConfig],
+  private[vertx] val codecClasses: Seq[Class[_]]) {
+  val readAdapters: Seq[VertxReadAdapterConfig] =
+    configurations.collect({ case c: VertxReadAdapterConfig => c })
+
+  val writeAdapters: Seq[VertxWriteAdapterConfig] =
+    configurations.collect({ case c: VertxWriteAdapterConfig => c })
+
+  def addAdapter(first: VertxAdapterConfig, rest: VertxAdapterConfig*): VertxAdapterSystemConfig =
```
krasserm: Shouldn't a method with name `addAdapter` have only a single `VertxAdapterConfig` parameter? All usage examples also don't use the `rest` vararg parameter.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
```diff
@@ -0,0 +1,68 @@
...
+    addAdapters(first +: rest.toVector)
```
krasserm: `copy(onfigurations = onfigurations :+ first)` or is ordering significant?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
```diff
@@ -0,0 +1,68 @@
...
+    else
+      invalid.map(c => s"Ambigious definition for adapter with id '${c._1}' given. An id may only be used once.")
+        .toVector
+        .left
+  }
+}
+
+class VertxAdapterSystemConfig(private[vertx] val configurations: Seq[VertxAdapterConfig],
+  private[vertx] val codecClasses: Seq[Class[_]]) {
+  val readAdapters: Seq[VertxReadAdapterConfig] =
```
krasserm: Why `private[vertx]`?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
```diff
@@ -0,0 +1,68 @@
...
+      configs.right
+    else
+      invalid.map(c => s"Ambigious definition for adapter with id '${c._1}' given. An id may only be used once.")
+        .toVector
+        .left
+  }
+}
+
+class VertxAdapterSystemConfig(private[vertx] val configurations: Seq[VertxAdapterConfig],
+  private[vertx] val codecClasses: Seq[Class[_]]) {
```
krasserm: Why `private[vertx]`?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  private def validateConfigurations(configs: Seq[VertxAdapterConfig]): \/[Seq[String], Seq[VertxAdapterConfig]] = {
+    val invalid = configs.groupBy(_.id).filter(c => c._2.size > 1)
+
+    if (invalid.isEmpty)
+      configs.right
+    else
+      invalid.map(c => s"Ambigious definition for adapter with id '${c._1}' given. An id may only be used once.")
+        .toVector
+        .left
```
krasserm: This generates n error messages that only differ in `${c._1}`. Why not just collecting keys with more than one config and create a single error message? What about using just `Either` instead of Scalaz's `\/`?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterConfig.scala
```diff
@@ -0,0 +1,102 @@
...
+}
+sealed trait VertxReadAdapterConfig extends VertxAdapterConfig {
+  def endpointRouter: VertxEndpointRouter
+}
+
+case class VertxPublishAdapterConfig(id: String, log: ActorRef, endpointRouter: VertxEndpointRouter) extends VertxReadAdapterConfig
+case class VertxSendAdapterConfig(id: String, log: ActorRef, endpointRouter: VertxEndpointRouter, deliveryMode: DeliveryMode) extends VertxReadAdapterConfig
+case class VertxWriteAdapterConfig(id: String, log: ActorRef, endpoints: Seq[String], filter: PartialFunction[Any, Boolean]) extends VertxAdapterConfig
+
+sealed trait ConfirmationType {}
```
krasserm: `{}` can be removed

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterConfig.scala
```diff
@@ -0,0 +1,102 @@
...
+case object Single extends ConfirmationType
+case class Batch(size: Int) extends ConfirmationType
+
+sealed trait DeliveryMode {}
```
krasserm: `{}` can be removed

eventuate-adapter-vertx/src/it/scala/com/rbmhtechnology/eventuate/adapter/vertx/ActorStorage.scala
```diff
@@ -0,0 +1,54 @@
...
+package com.rbmhtechnology.eventuate.adapter.vertx
+
+import akka.pattern.ask
+import akka.testkit.{TestKit, TestProbe}
+import akka.util.Timeout
+import com.rbmhtechnology.eventuate.adapter.vertx.api.StorageProvider
+import org.scalatest.{BeforeAndAfterEach, Suite}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
```
krasserm: Some imports are not used here

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
cstub: You're right - I will rename the varags-version to addAdapter**s**

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/AkkaSerializationMessageCodec.scala
```diff
@@ -0,0 +1,77 @@
...
+  def apply(name: String)(implicit system: ActorSystem): MessageCodec[AnyRef, AnyRef] =
+    new AkkaSerializationMessageCodec(name)
+
+  def apply(clazz: Class[_])(implicit system: ActorSystem): MessageCodec[AnyRef, AnyRef] =
+    new AkkaSerializationMessageCodec(s"${AkkaSerializationMessageCodec.Name}-${clazz.getName}")
+}
+
+class AkkaSerializationMessageCodec(override val name: String)(implicit system: ActorSystem) extends MessageCodec[AnyRef, AnyRef] {
+
+  val serializer = PayloadSerializationExtension(system)
```
krasserm: You can avoid creating your own serialization extension by using Akka's `SerializationExtension` and defining a payload wrapper (preferrably using protobuf) so that you know the class when deserializing with `Serialization.deserialize[T](bytes: Array[Byte], clazz: Class[T])`.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/AkkaSerializationMessageCodec.scala
```diff
@@ -0,0 +1,77 @@
...
+  override def encodeToWire(buffer: Buffer, o: AnyRef): Unit = {
+    val payload = serializer.toBinary(o)
+    buffer.appendInt(payload.length)
+    buffer.appendBytes(payload)
+  }
+
+  override def decodeFromWire(pos: Int, buffer: Buffer): AnyRef = {
+    val payloadLength = buffer.getInt(pos)
+    val payload = buffer.getBytes(pos + Integer.BYTES, pos + Integer.BYTES + payloadLength)
+    serializer.fromBinary(payload).asInstanceOf[AnyRef]
```
krasserm: Low level position arithmetic can be avoided with a protobuf-defined payload wrapper.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
cstub: Nope, ordering is not significant in this case, but does appending the element have any advantages?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/AkkaSerializationMessageCodec.scala
```diff
@@ -0,0 +1,77 @@
...
+  override def createExtension(system: ExtendedActorSystem): PayloadSerializationExtension =
+    new PayloadSerializationExtension(system)
+
+  override def get(system: ActorSystem): PayloadSerializationExtension =
+    super.get(system)
+}
+
+class PayloadSerializationExtension(system: ExtendedActorSystem) extends Extension {
+
+  val serializer = new CommonSerializer(system)
```
krasserm: When implementing the previously mentioned approach, the `serializer` can be moved to `AkkaSerializationMessageCodec`

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
cstub: I didn't want to expose the constructor as part of the public API.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/AkkaSerializationMessageCodec.scala
krasserm: Was there a special reason not implementing like that?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
krasserm: Why not removing the `rest` parameter instead? They're never used in the example

cstub: you're right - the varargs version is not necessary, I will remove the parameter.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
krasserm: No, shoudn't make a difference

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
krasserm: ok

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/AkkaSerializationMessageCodec.scala
cstub: I wanted to reuse the CommonSerializer as it already uses protobuf to wrap the the payload, but  this implementation needen an `ExtendedActorSystem` which was only available within an `Extension`

krasserm: You can access an `ExtendedActorSystem` from `Serialization`.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/AkkaSerializationMessageCodec.scala
cstub: According to the official [Vert.x examples](https://github.com/vert-x3/vertx-examples/blob/master/core-examples/src/main/java/io/vertx/example/core/eventbus/messagecodec/util/CustomMessageCodec.java) the payload length must be stored in order to recover the part of the buffer that contains the payload in `decodeFromWire`.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxAdapterSystem.scala
```diff
@@ -0,0 +1,88 @@
...
+import com.rbmhtechnology.eventuate.adapter.vertx.japi.rx.{ StorageProvider => RxStorageProvider }
+import com.rbmhtechnology.eventuate.adapter.vertx.japi.{ StorageProvider => JStorageProvider, VertxAdapterSystemConfig => JVertxAdapterSystemConfig }
+import io.vertx.core.Vertx
+import io.vertx.rxjava.core.{ Vertx => RxVertx }
+
+import scala.collection.immutable.Seq
+
+object VertxAdapterSystem {
+
+  import VertxConverters._
```
krasserm: Unused import

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxAdapterSystem.scala
```diff
@@ -0,0 +1,88 @@
...
+
+class VertxAdapterSystem private[vertx] (config: VertxAdapterSystemConfig, vertx: Vertx, storageProvider: StorageProvider)(implicit system: ActorSystem) {
+
+  private def registerEventBusCodecs(): Unit = {
+    config.codecClasses.foreach(c =>
+      try {
+        vertx.eventBus().registerDefaultCodec(c.asInstanceOf[Class[AnyRef]], AkkaSerializationMessageCodec(c))
+      } catch {
+        case e: IllegalStateException =>
+          system.log.warning(s"An adapter codec for class ${c.getName} was configured, even though a default codec was already registered for this class.")
```
krasserm: Shouldn't we propagate this exception to the application?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxWriteRouter.scala
```diff
@@ -0,0 +1,64 @@
...
+  val consumers = configs
+    .flatMap { c => c.endpoints.distinct.map(e => Route(e, writers(c.id), c.filter)) }
+    .groupBy(_.sourceEndpoint)
+    .map { case (endpoint, routes) => installMessageConsumer(endpoint, routes) }
+
+  private def installMessageConsumer(endpoint: String, routes: Seq[Route]): MessageConsumer[Any] = {
+    val handler = (msg: Message[Any]) => {
+      routes.foreach { route =>
+        if (route.filter.applyOrElse(msg.body(), (_: Any) => false)) {
+          route.destinationLog ! PersistMessage(msg.body(), msg)
```
krasserm: Why not having only a single `Message` parameter for `PersistMessage` and `VertxWriter`extracts the event with `msg.body`?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxWriteRouter.scala
```diff
@@ -0,0 +1,64 @@
...
+object VertxWriteRouter {
+
+  case class Route(sourceEndpoint: String, destinationLog: ActorRef, filter: PartialFunction[Any, Boolean])
+
+  def props(config: Seq[VertxWriteAdapterConfig], vertx: Vertx): Props =
+    Props(new VertxWriteRouter(config, vertx))
+}
+
+class VertxWriteRouter(configs: Seq[VertxWriteAdapterConfig], vertx: Vertx) extends Actor {
+  import VertxHandlerConverters._
```
krasserm: Unused import

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxWriteRouter.scala
```diff
@@ -0,0 +1,64 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rbmhtechnology.eventuate.adapter.vertx
+
+import akka.actor.{ Actor, ActorRef, Props }
+import com.rbmhtechnology.eventuate.adapter.vertx.VertxWriter.PersistMessage
```
krasserm: Unused import

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxAdapterSystem.scala
cstub: As far as I remember we said that trying to override an existing message-codec-binding should not result in an error but in a warning.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxWriteRouter.scala
cstub: yes, a single parameter would suffice - I will change that

krasserm:
I'll skip now commenting further unused imports. Please check the whole PR for unused imports.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxWriteRouter.scala
```diff
@@ -0,0 +1,64 @@
...
+    .groupBy(_.sourceEndpoint)
+    .map { case (endpoint, routes) => installMessageConsumer(endpoint, routes) }
+
+  private def installMessageConsumer(endpoint: String, routes: Seq[Route]): MessageConsumer[Any] = {
+    val handler = (msg: Message[Any]) => {
+      routes.foreach { route =>
+        if (route.filter.applyOrElse(msg.body(), (_: Any) => false)) {
+          route.destinationLog ! PersistMessage(msg.body(), msg)
+        } else {
+          msg.reply(msg.body)
```
krasserm: Would an ACK be sufficient (here and in `VertxWriter`)?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxWriteRouter.scala
```diff
@@ -0,0 +1,64 @@
...
+    .toMap
+
+  val consumers = configs
+    .flatMap { c => c.endpoints.distinct.map(e => Route(e, writers(c.id), c.filter)) }
+    .groupBy(_.sourceEndpoint)
+    .map { case (endpoint, routes) => installMessageConsumer(endpoint, routes) }
+
+  private def installMessageConsumer(endpoint: String, routes: Seq[Route]): MessageConsumer[Any] = {
+    val handler = (msg: Message[Any]) => {
+      routes.foreach { route =>
```
krasserm: A message should have only a single reply. Allowing n routes here may erroneously lead to n replies. An endpoint should be restricted to publish to exactly one log.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxWriteRouter.scala
```diff
@@ -0,0 +1,64 @@
...
+
+import akka.actor.{ Actor, ActorRef, Props }
+import com.rbmhtechnology.eventuate.adapter.vertx.VertxWriter.PersistMessage
+import com.rbmhtechnology.eventuate.adapter.vertx.api.VertxWriteAdapterConfig
+import io.vertx.core.Vertx
+import io.vertx.core.eventbus.{ Message, MessageConsumer }
+
+object VertxWriteRouter {
+
+  case class Route(sourceEndpoint: String, destinationLog: ActorRef, filter: PartialFunction[Any, Boolean])
```
krasserm: Rename `destinationLog` to `writer`?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
```diff
@@ -0,0 +1,134 @@
...
+
+  protected def eventBus(): EventBus =
+    vertx.eventBus()
+
+  def vertx: Vertx
+
+  protected def produce[A](address: String, msg: Any, deliveryOptions: DeliveryOptions, handler: Handler[AsyncResult[Message[A]]]): Unit
+
+  def produce(address: String, msg: Any): Unit = {
+    produce[Unit](address, msg, new DeliveryOptions(), ((ar: AsyncResult[Message[Unit]]) => {}).asVertxHandler)
```
krasserm: Equivalent `DeliveryOptions` and handler objects are created with every call. Define them in an outer scope? OTOH I'm not sure if the compiler is clever enough to detect that and optimize.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  def produce[A](address: String, msg: Any, timeout: FiniteDuration = 30.seconds)(implicit ec: ExecutionContext): Future[A] = {
```
krasserm: Make default timeout configurable?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
```diff
@@ -0,0 +1,134 @@
...
+    produce(address, msg, new DeliveryOptions().setSendTimeout(timeout.toMillis), promise.asVertxHandler)
+    promise.future.map(_.body)
+  }
+}
+
+trait MessagePublisher extends MessageProducer {
+  override protected def produce[A](address: String, msg: Any, deliveryOptions: DeliveryOptions, handler: Handler[AsyncResult[Message[A]]]): Unit = {
+    eventBus().publish(address, msg, deliveryOptions)
+    handler.handle(VertxFuture.succeededFuture(new MessageImpl[Any, A]()))
+  }
```
krasserm: From a design perspective, shouldn't a `MessagePublisher` rather offer a `publish` method without a `handler` parameter (and use a `produce` method to delegate to)?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
```diff
@@ -0,0 +1,134 @@
...
+}
+
+trait MessageSender extends MessageProducer {
+  override protected def produce[A](address: String, msg: Any, deliveryOptions: DeliveryOptions, handler: Handler[AsyncResult[Message[A]]]): Unit = {
+    eventBus().send(address, msg, deliveryOptions, handler)
+  }
```
krasserm: `send` method here (see also https://github.com/RBMHTechnology/eventuate/pull/310/files#r79381579)

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
```diff
@@ -0,0 +1,134 @@
...
+
+trait MessageProducer {
+  import VertxHandlerConverters._
+
+  protected def eventBus(): EventBus =
+    vertx.eventBus()
+
+  def vertx: Vertx
+
+  protected def produce[A](address: String, msg: Any, deliveryOptions: DeliveryOptions, handler: Handler[AsyncResult[Message[A]]]): Unit
```
krasserm: You're trying to abstract over point-to-point and pub-sub messaging here. Does that make sense?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
cstub: Yes, but the problem with this approach would be, that the `MessagePublisher` and `MessageSender` could not be used indistinguishable, as done in the various reader-actors.

This way a reader only has to depend on a `MessageProducer` and the mixed-in implementation will decide if it is a `Publisher` or a `Sender`

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
cstub: In this case I would say yes, because from a  sender point-of-view a publish can be seen as a send with an immediate response.
This enables generic handling of message-sending in the `VertxReader`

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
```diff
@@ -0,0 +1,134 @@
...
+  }
+}
+
+trait MessageSender extends MessageProducer {
+  override protected def produce[A](address: String, msg: Any, deliveryOptions: DeliveryOptions, handler: Handler[AsyncResult[Message[A]]]): Unit = {
+    eventBus().send(address, msg, deliveryOptions, handler)
+  }
+}
+
+trait MessageDelivery extends MessageProducer {
```
krasserm: For reasons of symmetry to the adapter configuration, `MessageDelivery` sub-traits should only extend the behavior of `MessageSender`. At-most-once semantics is implied for `MessagePublisher`. This is also related to https://github.com/RBMHTechnology/eventuate/pull/310/files#r79381579

krasserm:
Some thoughts regarding naming: shouldn't we rather use 

- *bus writer* instead of *vert.x reader* 
- *log writer* instead of *vert.x writer*

This would make declarations like

```scala
trait VertxReader[R, W] extends EventsourcedWriter[R, W] with ...
```

less confusing. WDYT?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
```diff
@@ -0,0 +1,134 @@
...
+    storageProvider.writeProgress(id, snr)
+
+  override def readProgress(id: String)(implicit executionContext: ExecutionContext): Future[Long] =
+    storageProvider.readProgress(id)
+
+  override def progress(result: Long): Long =
+    result
+}
+
+trait VertxReader[R, W] extends EventsourcedWriter[R, W] with MessageDelivery with ProgressStore[R, W] {
```
krasserm: See https://github.com/RBMHTechnology/eventuate/pull/310#issuecomment-247986683

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxSingleConfirmationSender.scala
```diff
@@ -0,0 +1,82 @@
...
+        .map(_ => Confirm(deliveryId))
+        .recover {
+          case err => DeliverFailed(envelope, deliveryId, err)
+        }
+        .pipeTo(self)
+
+    case Confirm(deliveryId) if unconfirmed.contains(deliveryId) =>
+      persistConfirmation(DeliveryConfirmed(), deliveryId) {
+        case Success(evt) =>
+        case Failure(err) => throw new RuntimeException(err)
```
krasserm: Just `log.error` here? The delivery will be retried and the confirmation will be received again.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
krasserm: In this case you may want to abstract over message delivery somewhere else (e.g. on reader-level) rather than here. Completing an asynchronous handler where the return from a from a method call is sufficient is a code/design smell IMO.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
krasserm: I consider a response semantically different from a send-acknowledgement. Responses are communicated via an asynchronous result handler, send/publish acknowledgements by returning from the send/publish method.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxReader.scala
cstub: I've removed the default timeout altogether because it is not used anyway.

eventuate-adapter-vertx/build.sbt
```diff
@@ -0,0 +1,2 @@
+// make protobuf definitions of core module available in this module (needed for compilation with protoc)
+includePaths in protobufConfig ++= (sourceDirectories in (LocalProject("core"), protobufConfig)).value
```
krasserm: Do we need that file? Seems that you don't use protobuf in the vert.x adapter.

krasserm: Also, in the main build.sbt file you can remove `.settings(protocSettings: _*)`

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterConfig.scala
```diff
@@ -0,0 +1,102 @@
...
+  def id: String
+  def log: ActorRef
+}
+
+sealed trait VertxDispatcherConfig extends VertxAdapterConfig {
+  def endpointRouter: EndpointRouter
+}
+
+case class VertxPublisherConfig(id: String, log: ActorRef, endpointRouter: EndpointRouter) extends VertxDispatcherConfig
+case class VertxSenderConfig(id: String, log: ActorRef, endpointRouter: EndpointRouter, deliveryMode: DeliveryMode) extends VertxDispatcherConfig
```
krasserm: Hmmm, there is a `LogDispatcherConfig`. For symmetry reasons, why not renaming:

- `VertxDispatcherConfig` -> `BusDispatcherConfig`
- `VertxPublisherConfig` -> `BusPublisherConfig`
- `VertxSenderConfig` -> `BusSenderConfig`

krasserm: Why have you renamed `...Writer...` to `...Dispatcher...`? I find `...Writer...` much better.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxEventDispatcher.scala
```diff
@@ -0,0 +1,105 @@
...
+trait VertxEventProducer {
+  def vertx: Vertx
+}
+
+trait VertxEventPublisher extends VertxEventProducer with EventPublisher {
+  override def publish(address: String, evt: Any): Unit =
+    vertx.eventBus().publish(address, evt)
+}
+
+trait VertxEventSender extends VertxEventPublisher with EventSender {
```
krasserm: extends `VertxEventProducer`?

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/VertxSender.scala
```diff
@@ -0,0 +1,31 @@
...
+import io.vertx.core.Vertx
+
+private[eventuate] object VertxSender {
+  def props(id: String, eventLog: ActorRef, endpointRouter: EndpointRouter, vertx: Vertx, storageProvider: StorageProvider): Props =
+    Props(new VertxSender(id, eventLog, endpointRouter, vertx, storageProvider))
+      .withDispatcher("eventuate.log.dispatchers.write-dispatcher")
+}
+
+private[eventuate] class VertxSender(val id: String, val eventLog: ActorRef, val endpointRouter: EndpointRouter, val vertx: Vertx, val storageProvider: StorageProvider)
+  extends VertxEventDispatcher[Long, Long] with AtMostOnceDelivery with VertxEventSender with SequenceNumberProgressStore {
```
krasserm: `AtMostOnceDelivery` causes the `VertxEventDispatcher` to **publish** on the Vert.x bus. The `VertxEventSender.send()` is never used here. Furthermore, when applying my [previous change](https://github.com/RBMHTechnology/eventuate/pull/310/files#r80664207) you anyway must use `VertxEventPublisher` here for making the code compile. 

The problem is that the `*Delivery` traits are not independent of the `VertxEvent*` traits. Maybe you design the `*Delivery` traits in a way that they are only applicable to `VertxEventSender`. At-least-once delivery for `VertxEventPublisher` would/should be implied then.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/VertxAdapterSystemConfig.scala
```diff
@@ -0,0 +1,85 @@
...
+    } yield configs
+  }
+
+  private def validateConfigurationIds(configs: Seq[VertxAdapterConfig]): Either[String, Seq[VertxAdapterConfig]] = {
+    val invalid = configs.groupBy(_.id).filter(c => c._2.size > 1)
+
+    if (invalid.isEmpty)
+      Right(configs)
+    else
+      Left(s"Ambigious definition for adapter(s) [${invalid.keys.map(id => s"'$id'").mkString(", ")}] given. " +
```
krasserm: Ambig\*u\*ous

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/LogWriter.scala
```diff
@@ -0,0 +1,50 @@
...
+
+private[vertx] object LogWriter {
+
+  case class PersistMessage(message: Message[Any])
+
+  def props(id: String, eventLog: ActorRef): Props =
+    Props(new LogWriter(id, eventLog))
+}
+
+private[vertx] class LogWriter(val id: String, val eventLog: ActorRef) extends EventsourcedActor {
```
krasserm: This actor might take a long time to replay events. Either safe periodically empty snapshots to reduce recovery time or use [this new feature](http://rbmhtechnology.github.io/eventuate/reference/event-sourcing.html#recovery-using-an-application-defined-log-sequence-number) to start replay from a custom sequence number. In the latter case, the sequence number must be managed elsewhere.

krasserm:
Awesome contribution @cstub :+1:  LGTM after adding API and user documentation.

eventuate-adapter-vertx/src/main/scala/com/rbmhtechnology/eventuate/adapter/vertx/api/EventProducerConfig.scala
```diff
@@ -0,0 +1,154 @@
...
+
+  override def as(id: String): EventProducerConfig =
+    VertxSenderConfig(id, log, endpointRouter, deliveryMode)
+}
+
+class LogProducerConfigFactory(endpoints: Set[String]) {
+
+  /**
+   * Sets the destination event log all received events are persisted to.
+   * An optional filter function can be supplied. Only events passing the function are persisted.
```
krasserm: passing the *filter* ...

src/sphinx/reference/adapters.rst
```diff
@@ -54,18 +54,221 @@ For persisting the stream processing progress, an application should store the l
...
+*Vert.x producers* are created by using the ``EventProducer.fromLog`` method. Applications define the source log, the delivery method and an arbitrary amount of event bus endpoints, the events will be delivered to.
+Vert.x producers support routing of events based on the content of the event.
+
+.. tabbed-code::
+   .. includecode:: ../../../eventuate-example-vertx/src/main/scala/com/rbmhtechnology/docs/vertx/Documentation.scala
+      :snippet: vertx-event-producer
+   .. includecode:: ../../../eventuate-example-vertx/src/main/java/com/rbmhtechnology/docs/vertx/japi/Documentation.java
+      :snippet: vertx-event-producer
+
+*Log producers* are created by using the ``EventProducer.fromEndpoints`` method. Multiple event bus endpoints can be defined, which are used to consume events from the event bus and persist the same events to the given event log. Events can be filtered by defining a filter function.
```
krasserm: Either include an example of a filter function or mention that it is not shown in the example

src/sphinx/reference/adapters.rst
```diff
@@ -54,18 +54,221 @@ For persisting the stream processing progress, an application should store the l
...
+For events to be written to an event log, applications send events to the specified endpoints configured for a log producer. The log producer consumes all events from these endpoints and persists the events to the configured event log.
+
+.. tabbed-code::
+   .. includecode:: ../../../eventuate-example-vertx/src/main/scala/com/rbmhtechnology/docs/vertx/Documentation.scala
+      :snippet: event-processing-log-producer
+   .. includecode:: ../../../eventuate-example-vertx/src/main/java/com/rbmhtechnology/docs/vertx/japi/Documentation.java
+      :snippet: event-processing-log-producer
+
+.. note::
+   Event processing in event bus handlers should be performed idempotent because a Vert.x producer may deliver the same event multiple times under certain conditions. Events may be redelivered after a restart of a producer if it was not able to successfully persist its replication progress on shutdown.
```
krasserm: > ... its replication progress on shutdown

its read progress on shutdown (or crash)

src/sphinx/reference/adapters.rst
```diff
@@ -54,18 +54,221 @@ For persisting the stream processing progress, an application should store the l
...
+Applications invoke the ``start`` method of the system to initialize the registered event producers and start event exchange with the ``Vert.x`` instance.
+
+.. tabbed-code::
+   .. includecode:: ../../../eventuate-example-vertx/src/main/scala/com/rbmhtechnology/docs/vertx/Documentation.scala
+      :snippet: adapter-example
+   .. includecode:: ../../../eventuate-example-vertx/src/main/java/com/rbmhtechnology/docs/vertx/japi/Documentation.java
+      :snippet: adapter-example
+
+.. warning::
+   The ``start`` method should only be called once all handlers on the event bus have been registered. Failing to do so may lead to loss of events because an producer might try to deliver events to an event bus endpoint which has not yet an event handler assigned to it.
```
krasserm: > should only be called once all handlers

called after all handlers

> an producer

a producer

src/sphinx/reference/adapters.rst
krasserm: ... or include a hint that event producers are covered in more detail further below (linking to the detail section)

src/sphinx/reference/adapters.rst
```diff
@@ -54,18 +54,221 @@ For persisting the stream processing progress, an application should store the l
...
+
+The following sections contain a detailed description of the different kinds of event producers.
+
+Vert.x Publish Event Producer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A *Publish Event Producer* publishes events from an event log to *multiple* subscribers on the event bus. Events are delivered to specific endpoints defined in the configuration of the producer. A producer can route events to different event bus endpoints based on the content of the event. Event routing is enabled by supplying a partial function which maps events to event bus endpoints. If no endpoint is returned by the function the event will not be processed.
```
krasserm: > If no endpoint is returned by the function

if the partial function is not defined at the event ...

src/sphinx/reference/adapters.rst
```diff
@@ -54,18 +54,221 @@ For persisting the stream processing progress, an application should store the l
...
+
+.. tabbed-code::
+   .. includecode:: ../../../eventuate-example-vertx/src/main/scala/com/rbmhtechnology/docs/vertx/Documentation.scala
+      :snippet: vertx-publish-producer
+   .. includecode:: ../../../eventuate-example-vertx/src/main/java/com/rbmhtechnology/docs/vertx/japi/Documentation.java
+      :snippet: vertx-publish-producer
+
+Event delivery is always performed with *At-Most-Once* delivery semantics, so no guarantees about the successful delivery of events can be made.
```
krasserm: Event publishing is performed with *At-Most-Once* delivery semantics.

src/sphinx/reference/adapters.rst
```diff
@@ -54,18 +54,221 @@ For persisting the stream processing progress, an application should store the l
...
+
+Applications consume events by registering an event bus handler at the configured endpoints.
```
krasserm: Why are these called *event bus handler* and not just *event handler*?

src/sphinx/reference/adapters.rst
```diff
@@ -54,18 +54,221 @@ For persisting the stream processing progress, an application should store the l
...
+
+.. tabbed-code::
+   .. includecode:: ../../../eventuate-example-vertx/src/main/scala/com/rbmhtechnology/docs/vertx/Documentation.scala
+      :snippet: event-processing-vertx-producer
+   .. includecode:: ../../../eventuate-example-vertx/src/main/java/com/rbmhtechnology/docs/vertx/japi/Documentation.java
+      :snippet: event-processing-vertx-producer
+
+Replication progress from the source event log is tracked by persisting the ``localSequenceNr`` of the latest delivered event to the ``StorageProvider`` supplied to the ``VertxAdapter``. After delivery of one or multiple events the replication progress is persisted. The producer continues delivery of events from the latest known ``localSequenceNr`` once the it is started.
```
krasserm: > Replication progress

read progress

> latest delivered event

The latest *sent* event

Delivery usually means delivery to the consumer but in this context you can only reason about successful sending/publishing to the event bus. So other occurrences of *deliver...* should be replaced as well here.

src/sphinx/reference/adapters.rst
```diff
@@ -54,18 +54,221 @@ For persisting the stream processing progress, an application should store the l
...
+
+Point-to-point event producers support both *At-Most-Once* and *At-Least-Once* delivery semantics. If not specified otherwise *At-Most-Once* delivery is chosen. *At-Least-Once* delivery is enabled by configuring the adapter accordingly.
+
+.. tabbed-code::
+   .. includecode:: ../../../eventuate-example-vertx/src/main/scala/com/rbmhtechnology/docs/vertx/Documentation.scala
+      :snippet: vertx-ptp-producer-at-least-once
+   .. includecode:: ../../../eventuate-example-vertx/src/main/java/com/rbmhtechnology/docs/vertx/japi/Documentation.java
+      :snippet: vertx-ptp-producer-at-least-once
+
+Events sent by a point-to-point event producer are received by registering an event handler on the event bus.
```
krasserm: Ha! Here's the *event handler* 😃

src/sphinx/reference/adapters.rst
```diff
@@ -54,18 +54,221 @@ For persisting the stream processing progress, an application should store the l
...
+
+Using *At-Least-Once* delivery semantics, every event must be confirmed by the receiver. Unconfirmed events are redelivered until a confirmation was received by the adapter. Event handlers signal event confirmation by replying to the event bus message with an arbitrary value.
```
krasserm: > Event handlers signal event confirmation

Event handlers confirm event delivery

> with an arbitrary value

Why that? Why not ACK?

src/sphinx/reference/adapters.rst
```diff
@@ -54,18 +54,221 @@ For persisting the stream processing progress, an application should store the l
...
+      :snippet: vertx-ptp-producer-handler
+
+Event confirmations are persisted on a per-event basis or in batches of configurable size.
+
+- **Per-event confirmations**:
+  Using per-event confirmations, every confirmation received by the adapter is persisted to the source event log. Confirmation events are not delivered to any event bus handlers but will increase the size of the source event log. With this confirmation mode events will not be redelivered once an event confirmation has been received.
+
+- **Batch event confirmations**:
+  Using batch confirmations, events are delivered in batches where the next batch is only delivered once all events of the previous batch have been confirmed. Batches containing events which have not been confirmed are redelivered as a whole, resulting in redelivery of all events of the same batch. This approach leads to modest storage requirements as no dedicated confirmation information has to be tracked. Using this confirmation mode, events may be redelivered multiple times even though a confirmation has already been received.
```
krasserm: > no dedicated confirmation information

no individual per-event confirmation (?)

krasserm:
@cstub very good documentation! After having fixed the build failures, please squash the commits. Ready for merge to master I think 😃.

src/sphinx/reference/adapters.rst
cstub: The adapter does not process the returned value, so the handler can send any value. 
We could define a case object `ACK` which can be sent? ok for you?

PR #322, opened 2016-10-11
Differentiate between local/remote timeouts

volkerstampa:
use dedicated timeouts for ask operations depending on the
type of operation:

- read from local event log
- write to local event log
- remote operation


krasserm:
LGTM

PR #333, opened 2016-10-17
Minor CRDT API and implementation enhancements

krasserm:
- rename update to effect (matches academic literature)
- rename set operation to assign (for Register CRDTs)
- needed for upcoming Eventuate CRDT article


volkerstampa:
LGTM

PR #336, opened 2016-10-20
Limit size of batches written by processor

### LINKED ISSUE #329: EventsourcedProcessor: write batch size exceeding partition size
Hi,

I just ran into the following exception with one of my `EventsourcedProcessor`s (a `StatefulProcessor` to be specific). During recovery the processor crashes with a write failure due to the write batch size exceeding the configured partition size (here `16000`).

```
Caused by: java.lang.IllegalArgumentException: requirement failed: write batch size (28611) must not be greater than maximum partition size (16000)
	at scala.Predef$.require(Predef.scala:224)
	at com.rbmhtechnology.eventuate.log.EventLog$.com$rbmhtechnology$eventuate$log$EventLog$$adjustSequenceNr(EventLog.scala:675)
	at com.rbmhtechnology.eventuate.log.EventLog$$anonfun$writeBatches$1.apply(EventLog.scala:558)
	at com.rbmhtechnology.eventuate.log.EventLog$$anonfun$writeBatches$1.apply(EventLog.scala:558)
	at scala.util.Try$.apply(Try.scala:192)
	at com.rbmhtechnology.eventuate.log.EventLog.writeBatches(EventLog.scala:558)
	at com.rbmhtechnology.eventuate.log.EventLog.com$rbmhtechnology$eventuate$log$EventLog$$processReplicationWrites(EventLog.scala:528)
	at com.rbmhtechnology.eventuate.log.EventLog$$anonfun$com$rbmhtechnology$eventuate$log$EventLog$$initialized$1.applyOrElse(EventLog.scala:433)
	at akka.actor.Actor$class.aroundReceive(Actor.scala:484)
	at com.rbmhtechnology.eventuate.log.EventLog.aroundReceive(EventLog.scala:247)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:526)
	at akka.actor.ActorCell.invoke(ActorCell.scala:495)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:257)
	at akka.dispatch.Mailbox.run(Mailbox.scala:224)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

I think the `EventsourcedProcessor`s `write()` requests a `ReplicationWrite` without respecting any batch sizes at all - not sure if that's the real issue though.

Cheers,
Gregor

krasserm:
- write events in multiple batches sequentially if the number of generated events during processing since the last write is greater than settings.writeBatchSize
- write events in a single batch if the number of generated events during processing since the last write is less than or equal to settings.writeBatchSize
- stop the processor if the number of generated events during processing of a single event is greater than settings.writeBatchSize
- replayBatchSize reset to default value (was settings.writeBatchSize)

- closes #329

### DIFF
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedProcessor.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedProcessor.scala
@@ -27,6 +27,7 @@ import com.typesafe.config.Config
 import scala.collection.immutable.Seq
 import scala.concurrent._
 import scala.concurrent.duration._
+import scala.util._
 
 private class EventsourcedProcessorSettings(config: Config) {
   val readTimeout =
@@ -73,7 +74,7 @@ object EventsourcedProcessor {
  *
  * @see [[StatefulProcessor]]
  */
-trait EventsourcedProcessor extends EventsourcedWriter[Long, Long] {
+trait EventsourcedProcessor extends EventsourcedWriter[Long, Long] with ActorLogging {
   import ReplicationProtocol._
   import context.dispatcher
 
@@ -84,7 +85,7 @@ trait EventsourcedProcessor extends EventsourcedWriter[Long, Long] {
 
   private val settings = new EventsourcedProcessorSettings(context.system.settings.config)
 
-  private var processedEvents: Vector[DurableEvent] = Vector.empty
+  private var processedEvents: Vector[Seq[DurableEvent]] = Vector.empty
   private var processingProgress: Long = 0L
 
   /**
@@ -112,19 +113,22 @@ trait EventsourcedProcessor extends EventsourcedWriter[Long, Long] {
     case payload if processEvent.isDefinedAt(payload) =>
       val currentProcessedEvents = processEvent(payload)
       if (lastSequenceNr > processingProgress)
-        processedEvents = currentProcessedEvents.map(createEvent(_, lastHandledEvent.customDestinationAggregateIds)).foldLeft(processedEvents)(_ :+ _)
+        processedEvents = processedEvents :+ currentProcessedEvents.map(createEvent(_, lastHandledEvent.customDestinationAggregateIds))
   }
 
   /**
    * Asynchronously writes processed events that have been collected since the last write together
-   * with the current processing progress.
+   * with the current processing progress. If the number of processed events since the last write
+   * is greater than the configured `eventuate.log.write-batch-size` multiple batches with a size
+   * less than or equal to `eventuate.log.write-batch-size` will be written sequentially. When
+   * splitting into multiple batches it is guaranteed that the processing result of a single event
+   * is part of the same batch i.e. it is guaranteed that the processing result of a single event
+   * is written atomically. If the processing result of a single event is larger than
+   * `eventuate.log.write-batch-size`, `write` fails with a [[EventsourcedProcessingResultTooLargeException]].
    */
   override final def write(): Future[Long] =
     if (lastSequenceNr > processingProgress) {
-      val result = targetEventLog.ask(ReplicationWrite(processedEvents, lastSequenceNr, id, VectorTime.Zero))(Timeout(writeTimeout)).flatMap {
-        case s: ReplicationWriteSuccess => Future.successful(s.storedReplicationProgress)
-        case f: ReplicationWriteFailure => Future.failed(f.cause)
-      }
+      val result = writeBatches(processedEvents, processingProgress, lastSequenceNr)
       processedEvents = Vector.empty
       result
     } else Future.successful(processingProgress)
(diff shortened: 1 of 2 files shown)


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedProcessor.scala
```diff
@@ -186,9 +196,52 @@ trait EventsourcedProcessor extends EventsourcedWriter[Long, Long] {
...
+    var num = 0
+    // Ensure that a single processEvent result
+    // is not partitioned across write batches.
+    events.span { w =>
+      val size = w.size
+      if (size > settings.writeBatchSize) {
+        throw new EventsourcedProcessingResultTooLargeException(size, settings.writeBatchSize)
+      } else {
+        num += size
+        num <= settings.writeBatchSize || num == size
```
volkerstampa: I would assume the ` || num == size` can be removed.

How can `num == size` be true, if `num <= settings.writeBatchSize` is false? if `num == size` and `num > settings.writeBatchSize` => `size > settings.writeBatchSize` but this causes the exception.

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedProcessor.scala
```diff
@@ -186,9 +196,52 @@ trait EventsourcedProcessor extends EventsourcedWriter[Long, Long] {
...
+      }
+    }
+  }
+
+  private def writeBatches(events: Vector[Seq[DurableEvent]], previousProgress: Long, currentProgress: Long): Future[Long] = {
```
volkerstampa: If writing a 2nd batch fails the progress is kept at `previousProgress` and thus when the actor restarts the 1st batch is written again, right? This works as writing a batch is idempotent, correct?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedProcessor.scala
krasserm: Right, this was a left-over from a version when the condition `(size > settings.writeBatchSize)` wasn't there yet. It should be there if we allowed batches larger than `writeBatchSize`

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedProcessor.scala
krasserm: Correct.

volkerstampa:
I know it is already merged, but I thought about this again and came up with this alternative implementation. It collects the batches in `onEvent` which makes `writeBatches` simpler:


```scala
  override final def onEvent = {
    case payload if processEvent.isDefinedAt(payload) =>
      val currentProcessedEvents = processEvent(payload)
      if (lastSequenceNr > processingProgress) {
        val durableEvents = currentProcessedEvents.map(createEvent(_, lastHandledEvent.customDestinationAggregateIds))
        processedEvents = processedEvents.lastOption match {
          case Some(last) if last.size <= settings.writeBatchSize - durableEvents.size =>
            processedEvents.dropRight(1) :+ (last ++ durableEvents)
          case _ =>
            processedEvents :+ durableEvents
        }
      }
  }
}
```
```scala
  private def writeBatches(events: Vector[Seq[DurableEvent]], previousProgress: Long, currentProgress: Long): Future[Long] = {
    events.dropRight(1)
      .foldLeft(Future.successful(previousProgress))((future, batch) => future.flatMap(writeBatch(batch, _)))
      .flatMap(_ => events.lastOption.map(writeBatch(_, currentProgress)).getOrElse(Future.successful(currentProgress)))
  }
```
No handling of exceeded batch size yet, but should be easy to add.

Just an idea...

krasserm:
Doesn't that just move complexity from `writeBatches` to `onEvent` but keeps the overall complexity the same? Furthermore, it adds more work on the event dispatching/processing thread. I'd rather propose to keep making batches of right size within `writeBatches` but (in contrast to the current solution) do that asynchronously i.e. instead of running the first [`splitBatches`](https://github.com/RBMHTechnology/eventuate/blob/master/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedProcessor.scala#L216) synchronously, do it asynchronously within a `Future`. This would not only allow for better concurrency of processing and batching/writing but also avoid explicit usage of `Future.successful` and `Future.failed` within `writeBatches`.

krasserm:
@volkerstampa I just implemented [that](https://github.com/RBMHTechnology/eventuate/pull/336#issuecomment-255578198) in #341. Let's continue the discussion there.

PR #340, opened 2016-10-21
Drop support for remote ReplicationFilter

### LINKED ISSUE #297: Deprecate remote replication filters

volkerstampa:
- remove from ReplicationConnection API
- keep in ReplicationRead
- keep serialization for backwards compatibility

- closes #297


volkerstampa:
I actually propose to merge this after the 0.8 release. Otherwise if upgrading an Eventuate based application from 0.7 to 0.8 one might have to do an atomic upgrade. If there are for example two locations A and B and with 0.7 B had a remote replication filter that is replaced by a local replication filter (implemented in A) with 0.8 and B is upgraded first, there is no filtering in place until A is upgraded as well. If we keep remote replication filters, B could be upgraded (with remote replication filter still in place). Afterwards A can be upgraded introducing local replication filters so that with the next upgrade the remote replication filter of B can be dropped.

krasserm:
Make sense. Let's postpone.

PR #345, opened 2016-10-29
Publish integration & default test artifacts

dispalt:
Add this step to the core project only giving other projects access to the `EventLogSpec` and friends.

Issue #96, opened 2015-08-06
Storage plugin test kit


krasserm:

krasserm:
A compliance testkit for implementations of #95 and #144.

PR #348, opened 2016-10-31
Publish test and integrationTest artifacts for the core project

dispalt:
Add this step to the core project only giving other projects access to the EventLogSpec and friends.

Supersedes #345 

Also, this is not a public API, that is going to be covered in #96, but this bridges a gap.


krasserm:
When I'm locally running your changes I'm getting:

```
build.sbt:33: error: overloaded method value settings with alternatives:
  (ss: sbt.Def.Setting[_]*)sbt.Project <and>
  => Seq[sbt.Def.Setting[_]]
 cannot be applied to (sbt.Def.SettingsDefinition)
  .settings(publishIntegrationTestSettings: _*)
   ^
sbt.compiler.EvalException: Type error in expression
	at sbt.compiler.Eval.checkError(Eval.scala:384)
	at sbt.compiler.Eval.compileAndLoad(Eval.scala:183)
	at sbt.compiler.Eval.evalCommon(Eval.scala:152)
	at sbt.compiler.Eval.evalDefinitions(Eval.scala:122)
	at sbt.EvaluateConfigurations$.evaluateDefinitions(EvaluateConfigurations.scala:254)
    ...
```

I'm able to fix that using a `flatMap` over package keys. 

```scala
  lazy val publishIntegrationTestSettings = Seq(
      publishArtifact in Test := true
  ) ++ Classpaths.defaultPackageKeys.flatMap(tk =>
    addArtifact(artifact in (IntegrationTest, tk), tk in IntegrationTest)
  )
```

No idea why this is successfully running on Travis CI. Any ideas?

krasserm:
Just found that this is an issue of the installed sbt version. Had 0.13.6, now works with 0.13.13. Maybe we should still use `flatMap` to have it working with older versions. WDYT?

dispalt:
Yeah that works for me ill push a rebased change

krasserm:
@dispalt I checked again, a `flatMap` is correct because the result can be implicitly converted to `Seq[Setting[_]]` like all other settings:

```scala
  lazy val integrationTestPublishSettings: Seq[Setting[_]] = Seq(
    publishArtifact in Test := true
  ) ++ Classpaths.defaultPackageKeys.flatMap(tk =>
    addArtifact(artifact in (IntegrationTest, tk), tk in IntegrationTest)
  )
```

Here, I also changed `publishIntegrationTestSettings` to `integrationTestPublishSettings` to follow the naming pattern of other settings.

We don't need to support older sbt versions because with #353 we anyway enforce 0.13.13.

dispalt:
Updated to `flatMap` and also fixed the eol char.

krasserm:
Thanks for your contribution @dispalt

PR #349, opened 2016-10-31
Convert mac eol -> unix eol

dispalt:
Carriage Return only -> Line feed


krasserm:
Are you sure that this PR doesn't conflict with #348? I think it should be rebased on it or just add this commit to #348.

dispalt:
It does, but which ever one merges first Ill have to fix the other one.

Ill just add it to #348

PR #365, opened 2016-11-30
Limit number of update notifications sent

### LINKED ISSUE #364: Limit number of ReplicationDue sent per ReplicationRead
If `update-notifications` are [enabled](http://rbmhtechnology.github.io/eventuate/reference/event-log.html?highlight=update%20notification#update-notifications) and location A replicates events from location B, B sends `ReplicationDue` messages to A when new events are written to the log (and there is no pending replication read from A). B keeps sending these `ReplicationDue` messages even after A was shutdown resulting in corresponding WARN logs from akka-remote in B's application log.

To avoid spamming the application log in such circumstances the number of `ReplicationDue` sent to other locations on received replication read requests should be limited.

volkerstampa:
Do not send update notifications forever to avoid spamming the
application log due to akka remoting problems when partner was shutdown.

closes #364

### DIFF
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/log/NotificationChannel.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/log/NotificationChannel.scala
@@ -16,15 +16,32 @@
 
 package com.rbmhtechnology.eventuate.log
 
-import akka.actor.Actor
+import java.util.concurrent.TimeUnit
 
+import akka.actor.Actor
+import akka.actor.ActorRef
 import com.rbmhtechnology.eventuate._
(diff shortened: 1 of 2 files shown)


krasserm:
I'd prefer a solution where `NotificationChannel.registry` is a LRU cache where read requests are stored for a configurable amount of time, preferably `eventuate.log.replication.failure-detection-limit`. This would make failure detection behavior more symmetric i.e. both replicator and notification channel give up after the configured limit.

volkerstampa:
Wouldn't it make more sense to use `eventuate.log.replication.retry-delay` as expiration time? Once that has passed we anyways expect the next replication read request.

krasserm:
That's fine too I think

PR #367, opened 2016-12-01
Drop support for remote ReplicationFilter

### LINKED ISSUE #297: Deprecate remote replication filters

volkerstampa:
- remove from ReplicationConnection API
- keep in ReplicationRead
- keep serialization for backwards compatibility

- closes #297


src/sphinx/architecture.rst
```diff
@@ -77,12 +77,7 @@ The read load on local event logs and network bandwidth usage increases with the
...
-- :ref:`remote-replication-filters` can be defined per replication connection. They determine which events shall be replicated **from** other locations.
-
-Replication filters can be used to make event collaboration, especially event-driven communication, more directed\ [#]_. Without replication filters all events are replicated to all locations in a replication network.
+By default all events are replicated in a replicated event log. In many cases, however, a location only wants to replicate a subset of events to and from other locations. Restricting event replication to a subset of events can be done with *replication filters*. :ref:`replication-filters` can be defined per replication endpoint. They determine which events can be replicated **to** other locations. Replication filters can be used to make event collaboration, especially event-driven communication, more directed\ [#]_. Without replication filters all events are replicated to all locations in a replication network.
```
krasserm: I think we should a warning box, that filtered connections must not be used in a replication network cycle (maybe with a hint that future Eventuate versions will remove that constraint).

PR #368, opened 2016-12-02
toggle aggregate indexing on cassandra event log

kongo2002:
`CassandraEventLog.props` now supports an additional argument called
`aggregateIndexing` to toggle indexing operations on the event log. If
you don't use aggregate IDs disabling the indexing may significantly
reduce the pressure on the Cassandra database (especially with high
write loads).

### DIFF
--- a/eventuate-log-cassandra/src/it/scala/com/rbmhtechnology/eventuate/LocationSpecCassandra.scala
+++ b/eventuate-log-cassandra/src/it/scala/com/rbmhtechnology/eventuate/LocationSpecCassandra.scala
@@ -20,7 +20,6 @@ import java.io.File
 
 import akka.actor._
 import akka.testkit._
-
 import com.rbmhtechnology.eventuate.log._
 import com.rbmhtechnology.eventuate.log.cassandra._
 import com.rbmhtechnology.eventuate.log.cassandra.CassandraIndex.AggregateEvents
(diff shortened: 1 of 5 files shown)


krasserm:
Your solution only works if the application doesn't use event-sourced components with a defined `aggregateId` (as documented). This restriction is too severe in my opinion. We should still allow the application to use `aggregateId`s, however, when turning indexing off, an `aggregateId`-based replay should do a full log scan. This is not the case at the moment, because the index is initialized with the recovered clock which causes raw log reads from the clock's sequence number rather than from the specified sequence number for `aggregateId`-based replays.

krasserm:
LGTM after additional test(s)

kongo2002:
@krasserm I've added a test spec that execute the `EventlogSpec` tests using a Cassandra event log *without* `aggregateIndexing`

eventuate-log-cassandra/src/it/scala/com/rbmhtechnology/eventuate/LocationSpecCassandra.scala
```diff
@@ -98,6 +97,39 @@ object SingleLocationSpecCassandra {
   }
 }
 
+trait SingleLocationSpecNoIndexingCassandra extends SingleLocationSpec with LocationCleanupCassandra {
```
krasserm: This is unnecessary code duplication with `SingleLocationSpecCassandra`. Instead, try implementing `EventLogSpecCassandraNoIndexing` as

```scala
class EventLogSpecCassandraNoIndexing extends TestKit(ActorSystem("test", EventLogSpecCassandra.config)) with EventLogSpec with SingleLocationSpecCassandra {
  import SingleLocationSpecCassandra._

  override def logProps(logId: String, failureSpec: TestFailureSpec, indexProbe: ActorRef): Props =
    TestEventLog.props(logId, failureSpec, indexProbe, batching, aggregateIndexing = false)
}
```

and remove the `SingleLocationSpecNoIndexingCassandra` trait.

eventuate-log-cassandra/src/it/scala/com/rbmhtechnology/eventuate/log/EventLogSpecCassandra.scala
```diff
@@ -39,6 +39,8 @@ object EventLogSpecCassandra {
     """.stripMargin).withFallback(EventLogSpec.config)
 }
 
+class EventLogSpecNoIndexingCassandra extends TestKit(ActorSystem("test", EventLogSpecCassandra.config)) with EventLogSpec with SingleLocationSpecNoIndexingCassandra
```
krasserm: See previous comment

kongo2002:
@krasserm I just applied your suggestions and rebased onto `master`

krasserm:
Perfect, thanks!

PR #373, opened 2016-12-19
Allow EventsourcedProcessor to post-process events

### LINKED ISSUE #370: Allow EventsourcedProcessor to emit events with custom aggregation IDs
My question on the Gitter channel:

> I’m currently trying to create an EventsourcedProcessor that reads an event from event log A and writes a newly created event to event log B. I want the newly created event to be equipped with an aggregate ID. Unfortunately, AFAICS the event that will be sent to event log B can only contain the same “customDestinationAggregateIds” as the original event from event log A (which btw. doesn’t have and IMO shouldn’t have an aggregate ID).
Am I right here that the API does not permit to realise such a scenario? Is this intended?

... and @volkerstampa s answer

> You are right, the current API does not permit to define custom aggregate ids when processing events in an EventsourcedProcessor, however I do not see a reason why this should not be possible. Feel free to open an issue for this. PRs are also welcome :grinning:.

danbim:
- Closes #370 in which an EventsourcedProcessor was not able to define custom routing destination
  aggregate IDs for the events produced by `processEvent`
- Introduces method EventsourcedProcessor.postProcessDurableEvent: DurableEvent => DurableEvent
  which defaults to identity and allows manipulation of the DurableEvent instance created as a
  consequence of new events generated by `processEvent`


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedProcessor.scala
```diff
@@ -113,7 +120,9 @@ trait EventsourcedProcessor extends EventsourcedWriter[Long, Long] with ActorLog
     case payload if processEvent.isDefinedAt(payload) =>
       val currentProcessedEvents = processEvent(payload)
       if (lastSequenceNr > processingProgress)
-        processedEvents = processedEvents :+ currentProcessedEvents.map(createEvent(_, lastHandledEvent.customDestinationAggregateIds))
+        processedEvents = processedEvents :+ currentProcessedEvents.map { e =>
+          postProcessDurableEvent(createEvent(e, lastHandledEvent.customDestinationAggregateIds))
```
krasserm: This allows implementation classes to overwrite `emitterId`, `vectorTimestamp` and `processId` which must be prevented.

krasserm: Calling `postProcessDurableEvent` causes the creation of a new function object for each processed event. Function object creation should be done once and only function application should be repeated. Although we don't have a behavior stack here, this is similar to what an `EventsourcedView` does for `onCommand`, `onEvent` and `onSnapshot` handlers (see https://github.com/RBMHTechnology/eventuate/blob/master/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedView.scala#L113). Although not touched by your PR, the same problem exists for `processEvent` and should be changed accordingly. Such changes are a prerequisite for adding support for event processing [behavior changes](http://rbmhtechnology.github.io/eventuate/reference/event-sourcing.html#behavior-changes) later, if needed.

krasserm:
And please write tests for all changes/additions you make 😃

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/EventsourcedProcessor.scala
danbim: Taking a look at the fields in DurableEvent I'd say there are maybe more than these 3 fields you mentioned that shouldn't be changed:

```scala
case class DurableEvent(
  payload: Any,
  emitterId: String = DurableEvent.UndefinedEmittedId,
  emitterAggregateId: Option[String] = None,
  customDestinationAggregateIds: Set[String] = Set(),
  systemTimestamp: Long = 0L,
  vectorTimestamp: VectorTime = VectorTime.Zero,
  processId: String = DurableEvent.UndefinedLogId,
  localLogId: String = DurableEvent.UndefinedLogId,
  localSequenceNr: Long = DurableEvent.UndefinedSequenceNr,
  deliveryId: Option[String] = None,
  persistOnEventSequenceNr: Option[Long] = None)
```

Maybe I'm missing some use-cases here but I'm not exactly sure which other fields than `customDestinationAggregateIds` should be allowed to change in the first place:

- `payload` is "created" in `processEvent`, no need to allow to post-process it
- `emitterId` could be allowed to change, even though I don't have a use-case in mind now
- `systemTimestamp` should be set by the `EventLog` implementation as I now learned in the context of #370, right? If so, it shouldn't be allowed to be changed here, either.
- `vectorTimestamp` should not change as you said
- `processId` should not change as you said
- `localLogId` probably not !?
- `localSequenceNr` probably not !?
- `deliveryId` probably not !?
- `persistOnEventSequenceNr` probably not !?

The assumption of my initial implementation was "garbage in, garbage out", i.e. the implementor of the processor should know what he/she is doing. Therefore, I also didn't add a test because would be no real assumptions that can be verified (other than e.g. the use-case of changing the customDestinationAggregateId which I could obviously add).

WDYT? If nothing should be allowed to change other than `customDestinationAggregateIds` (and `emitterAggregateId`?) maybe it would be sensible to add something specific for that feature.

krasserm: Good arguments @danbim, thanks. We definitely need a better specification which `DurableEvent` fields should be allowed to change and which not. More importantly, both the actor-based and the stream-based event processors should enforce that specification consistently. I created a separate ticket for a common event processing model (#376) so that we can cover that later.

In context of this ticket, I'm fine with your initial proposal that is based on the "garbage in, garbage out" assumption. Let's keep it as an (undocumented) expert feature and document the work results of #376 later. It anyway closely matches what can be done with the stream-based processor at the moment.

Regarding the test, adding one that only changes `customDestinationAggregateIds` should be sufficient.

danbim:
Hey Martin!

Just added a test. I'm not sure about the "code style" but I think it covers what should be tested.

Cheers
Daniel

PR #386, opened 2017-03-22
Stylesheets

mslinn:
Most changes won't be noticeable for most people. However, people who are visually impaired, who use a different browser css setup will find these changes to their liking.

* The tightly cropped white border around `<code></code>` is gone. It added no benefit, and made code hard to read
* The side menu now shows `h2`-`h5` entries, instead of just `h2` entries. They are now indented appropriately and do not wrap, so they are shown one per line. Any x-overflow is clipped.
* `h2`-`h5` titles in the document now have slightly different sizes, so the document hierarchy is easier to follow
* Because the width of `<pre></pre>` blocks is so narrow, and the font used was so large, not many characters could be shown. The font is now slightly smaller (12 points, which is a standard size).
* Some spacings were adjusted; this should not be noticeable unless a careful A/B comparison is done.
* Paragraph text is now justified; the document looks more professional this way.
* Only hyperlinks are now shown in red. Previously, code in paragraphs was also shown in red, which made it difficult to know what was a link. Code in paragraphs is now shown in black.
* The text was previously shown in light grey against a white background. This made the document difficult to read for a visually impaired person. Text is now shown in a darker grey.
* Some backgrounds are now transparent, instead of white, because many visually impaired people use a different stylesheet that does not have a white background. Transparent backgrounds mean that the desired colors work better.

### DIFF
--- a/src/sphinx/_ext/tabbedcode.css
+++ b/src/sphinx/_ext/tabbedcode.css
@@ -1,7 +1,7 @@
 ul.tabbed-code-selector {
     margin-bottom: 0 !important;
     position: relative;
-    bottom: -22px;
+    bottom: -12px;
     list-style-type: none;
     display: inline-flex;
 }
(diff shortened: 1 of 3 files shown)


krasserm:
What's the status of CLA signing? I didn't receive a notification yet ...

mslinn:
I immediately signed. Some CLAs are objectionable, like the Lightbend 
CLA (I refused to sign that document). The RBMH CLA is just fine, no 
problems.

Mike

krasserm:
Manually merged after squash with https://github.com/RBMHTechnology/eventuate/commit/5a943627ae00b0e4f0a62b263e74fdfe6891fa95

PR #392, opened 2017-06-01
Upgrade to akka 2.5.2

### LINKED ISSUE #390: Upgrade to Akka 2.5
After an initial attempt it turns out that this requires significant changes to the Eventuate Java API

volkerstampa:
Akka 2.5.2 introduced the following incompatible changes in the java API
- ReceiveBuilder:
  - explicit create step required
  - builds AbstractActor.Receive instead of Actor.Receive
- Match.match: Changes in type parameters

To adjust to these changes AbstractEventsourcedView gets a new method
receiveBuilder (similar to the one in AbstractActor) and each setOn...
method is overloaded with a variant that accepts an
AbstractActor.Receive (instead of an Actor.Receive)

Closes #390

### DIFF
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/AbstractEventsourcedView.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/AbstractEventsourcedView.scala
@@ -18,9 +18,235 @@ package com.rbmhtechnology.eventuate
 
 import java.util.{ Optional => JOption }
 
-import akka.actor.{ Actor, ActorRef }
+import akka.actor.AbstractActor
+import akka.actor.Actor.Receive
+import akka.actor.ActorRef
+import akka.japi.pf.ReceiveBuilder
 
 import scala.compat.java8.OptionConverters._
+
+import scala.runtime.BoxedUnit
+
+private[eventuate] object ReceiveConverters {
+
+  implicit class ScalaReceiveConverter(receive: Receive) {
+    def asJava: AbstractActor.Receive =
+      new AbstractActor.Receive(receive.asInstanceOf[PartialFunction[Any, BoxedUnit]])
+  }
+
+  implicit class JavaReceiveConverter(receive: AbstractActor.Receive) {
+    def asScala: Receive =
+      receive.onMessage.asInstanceOf[Receive]
+  }
+}
+
+private[eventuate] object AbstractEventsourcedComponent {
+
+  import ReceiveConverters._
+
+  /**
+   * Java API of the actors [[com.rbmhtechnology.eventuate.BehaviorContext]].
+   *
+   * Provides a context for managing behaviors.
+   */
+  final class BehaviorContext private[eventuate] (context: com.rbmhtechnology.eventuate.BehaviorContext) {
+
+    /**
+     * The initial behavior.
+     */
+    def initial: AbstractActor.Receive =
+      context.initial.asJava
+
+    /**
+     * The current behavior.
+     */
+    def current: AbstractActor.Receive =
+      context.current.asJava
+
+    /**
+     * Sets the given `behavior` as `current` behavior. Replaces the `current` behavior on the behavior stack.
+     *
+     * @param behavior the behavior to set.
+     */
+    def become(behavior: AbstractActor.Receive): Unit =
+      become(behavior, replace = true)
+
+    /**
+     * Sets the given `behavior` as `current` behavior.
+     *
+     * @param behavior the behavior to set.
+     * @param replace if `true` (default) replaces the `current` behavior on the behavior stack, if `false`
+     *                pushes `behavior` on the behavior stack.
+     */
+    def become(behavior: AbstractActor.Receive, replace: Boolean): Unit =
+      context.become(behavior.asScala, replace)
+
+    /**
+     * Pops the current `behavior` from the behavior stack, making the previous behavior the `current` behavior.
+     * If the behavior stack contains only a single element, the `current` behavior is reverted to the `initial`
+     * behavior.
+     */
+    def unbecome(): Unit =
+      context.unbecome()
+  }
+}
+
+/**
+ * Java API for actors that implement any event-sourced component.
+ *
+ * @define akkaReceiveBuilder http://doc.akka.io/japi/akka/2.5.2/akka/japi/pf/ReceiveBuilder.html
+ */
+private[eventuate] abstract class AbstractEventsourcedComponent extends EventsourcedView with ConditionalRequests {
+
+  import ReceiveConverters._
+
+  /**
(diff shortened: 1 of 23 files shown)


eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/AbstractEventsourcedView.scala
```diff
@@ -151,6 +153,20 @@ abstract class AbstractEventsourcedView(val id: String, val eventLog: ActorRef)
...
+   *
+   * Supplied with a partial function that defines the actor's command handling behaviour.
+   * Use [[$akkaReceiveBuilder ReceiveBuilder]] to define the behaviour.
+   *
+   * If [[onCommand]] is implemented, the supplied behaviour is ignored.
+   *
+   * @param handler a function that defines this actor's command handling behaviour.
+   * @see [[EventsourcedView]]
+   */
+  final protected def setOnCommand(handler: AbstractActor.Receive): Unit =
```
cstub: To keep the existing Eventuate-API pattern but also align with the new Akka-Java API I would suggest to offer two variants of each handler - a `setOnXX` and a `createOnXX` (which mimics the new Akka-Java API).

This could look something like this
```
private var commandBehaviour: AbstractActor.Receive = AbstractActor.emptyBehavior

def createOnCommand(): AbstractActor.Receive =
  AbstractActor.emptyBehavior
  
final protected def setOnCommand(handler: AbstractActor.Receive): Unit =
  if (commandBehaviour == AbstractActor.emptyBehavior) commandBehaviour = handler
  else throw new IllegalStateException("Actor command behaviour has already been set with setOnCommand(...).  " +
      "Use commandContext.become(...) to change it.")

final override def onCommand =
  createOnCommand().orElse(commandBehaviour).onMessage.asInstanceOf[Receive]
```

This way the `setOnXXX` - API of Eventuate stays the same (facilitating the new `AbstractActor.Receive` ) but users could also decide to use the new Akka pattern instead of overriding `onXX` and returning the often confusing type of `PartialFunction<Object, BoxedUnit>`.

WDYT?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/AbstractEventsourcedView.scala
```diff
@@ -151,6 +153,20 @@ abstract class AbstractEventsourcedView(val id: String, val eventLog: ActorRef)
       "Use commandContext.become(...) to change it.")
```
cstub: To align with the new Akka-Java API of `AbstractActor` I would suggest to add the following methods:
```
def getContext(): AbstractActor.ActorContext
def getSelf(): ActorRef
def getSender(): ActorRef
```

volkerstampa: +1

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/AbstractEventsourcedView.scala
volkerstampa: +1 ... with some slight changes. See following commits ...

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/AbstractEventsourcedView.scala
```diff
@@ -68,109 +125,127 @@ private[eventuate] abstract class AbstractEventsourcedComponent extends Eventsou
...
-   *
-   * If [[onCommand]] is implemented, the supplied behaviour is ignored.
+   * Creates a new empty `ReceiveBuilder`. Mimics [[AbstractActor.receiveBuilder()]].
+   */
+  final def receiveBuilder(): ReceiveBuilder = ReceiveBuilder.create()
+
+  /**
+   * Returns this AbstractActor's ActorContext. Mimics [[AbstractActor.getContext()]].
+   */
+  final def getContext(): AbstractActor.ActorContext = context.asInstanceOf[AbstractActor.ActorContext]
```
volkerstampa: The counterparts of `getContext`, `getSelf` and `getSender` in `AbstractActor` are not final, so maybe we should keep them non-final here as well?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/AbstractEventsourcedView.scala
```diff
@@ -68,109 +125,127 @@ private[eventuate] abstract class AbstractEventsourcedComponent extends Eventsou
...
+   * Java API of [[EventsourcedView.recovering lastEmirecoveringterId]].
    *
-   * @param handler a function that defines this actor's command handling behaviour.
    * @see [[EventsourcedView]]
    */
-  final protected def setOnCommand(handler: AbstractActor.Receive): Unit =
-    if (commandBehaviour.isEmpty) commandBehaviour = Some(handler)
-    else throw new IllegalStateException("Actor command behaviour has already been set with setOnCommand(...).  " +
-      "Use commandContext.become(...) to change it.")
+  final def isRecovering: Boolean = recovering
```
volkerstampa: `recovering` still used by `VertxAdapterExample`

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/AbstractEventsourcedView.scala
volkerstampa: The counterparts of `getContext`, `getSelf` and `getSender` in `AbstractActor` are not final, maybe we should leave them non-final here as well?!?

cstub: Do you see any good reason why those accessors should not be final? Does it make sense to override them?

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/AbstractEventsourcedView.scala
cstub: thx - I will replace it with `isRecovering`

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/AbstractEventsourcedView.scala
volkerstampa: I do not see a specific use case atm for overriding however I am also not 100% sure if there isnt any. Question is, if the decision to make them non-final in `AbstractActor` was consciously or arbitrary. But for me it is also ok to keep them final.

calvinlfer:
FYI 2.5.6 is out, would it be possible to update to latest?

cstub:
@calvinlfer Thanks for the hint. We will update this branch to use Akka 2.5.6.

PR #394, opened 2017-06-28
Fix 393 inconsistent index on disaster recovery

volkerstampa:


volkerstampa:
@magro Thanks for the review. Will add a comment to the commit. The alternative I had seen was using ask when sending the `UpdateIndex` command from `updateIndex` to the index-actor. But that would have required to specify a timeout which I wanted to avoid as it is hard to define a reasonable default.

magro:
Yeah, that was what I was thinking as well, makes sense.

eventuate-log-cassandra/src/main/scala/com/rbmhtechnology/eventuate/log/cassandra/CassandraEventLog.scala
```diff
@@ -136,7 +137,7 @@ class CassandraEventLog(id: String, aggregateIndexing: Boolean) extends EventLog
 
   override def writeEventLogClockSnapshot(clock: EventLogClock): Future[Unit] = {
     import context.dispatcher
-    indexStore.writeEventLogClockSnapshotAsync(clock).map(_ => ())
+    updateIndex(clock).flatMap(_ => indexStore.writeEventLogClockSnapshotAsync(clock)).map(_ => ())
```
cstub: If `aggregateIndexing` is set to false in `updateIndex`, `indexStore.writeEventLogClockSnapshotAsync` will be invoked twice in succession with the same `clock` parameter.
I guess this won't happen and if so will not have any considerable performance implications.

eventuate-log-cassandra/src/main/scala/com/rbmhtechnology/eventuate/log/cassandra/CassandraEventLog.scala
```diff
@@ -183,25 +184,28 @@ class CassandraEventLog(id: String, aggregateIndexing: Boolean) extends EventLog
...
+    } else
+      // otherwise update the event log clock snapshot only
+      indexStore.writeEventLogClockSnapshotAsync(clock)
+  }
+
   override def unhandled(message: Any): Unit = message match {
     case u @ UpdateIndexSuccess(clock, _) =>
       indexSequenceNr = clock.sequenceNr
       onIndexEvent(u)
-    case u @ UpdateIndexFailure(_) =>
```
cstub: Is it no longer necessary to call `onIndexEvent` in case of a failure?

eventuate-log-cassandra/src/main/scala/com/rbmhtechnology/eventuate/log/cassandra/CassandraEventLog.scala
volkerstampa: What do you mean it won't happen? As you pointed out it will happen, if `aggregateIndexing` is true and disaster recovery is invoked, right? But I would tend to agree, this should not affect the performance considerably, so I would tend to keep it as it is to avoid additional complexity.

eventuate-log-cassandra/src/main/scala/com/rbmhtechnology/eventuate/log/cassandra/CassandraEventLog.scala
volkerstampa: It actually never was required as this is anyways only used in tests for inspecting the index progress and since there was no test inspecting failures I thought its ok to remove it here.

eventuate-log-cassandra/src/main/scala/com/rbmhtechnology/eventuate/log/cassandra/CassandraEventLog.scala
cstub: +1

eventuate-log-cassandra/src/main/scala/com/rbmhtechnology/eventuate/log/cassandra/CassandraEventLog.scala
cstub: +1

PR #397, opened 2017-09-21
RGArray CRDT

Horusiath:
Replicated Growable Arrays are operation-based CRDTs, which - unlike other types of CRDT collection - are able to maintain a fixed order of elements added and removed from them. I've found [this video](https://youtu.be/yCcWpzY8dIA?t=30m20s) to be extremely helpful in understanding them.

Essentially each element's position is recognized by `Position(order: Int, emitterId: String)` object, where `order` is a monotonically incremented value, each time new element is added by current emitter. *This is the part I'm struggling with, as I don't know how to expose current emitterId from CRDTService level.*

Deleted elements are not removed from the data structure (as this would result in information loss, causing unsolvable conflicts in concurrent updates), they are tombstoned instead.

Core `RGArray` API looks as follows:

```scala
case class Position(order: Int, emitterId: String) extends Ordered[Position]
case class Vertex[A](value: A, pos: Position, isTombstoned: Boolean = false)
case class RGArray[A](vertices: Vector[Vertex[A]] = Vector.empty[Vertex[A]]) {
  /** Replicated Growable Array represents an indexable sequence of elements */
  def value: Vector[A]
  /** 
    * Inserts an element as the first element of RGArray. `newPos` is used
    * as an index assigned to that element.
    */
  def insertRight(value: A, newPos: Position): RGArray[A]
  /** 
    * Inserts an element on the right side of a given index within current RGArray. 
    * `newPos` is used as an index assigned to that element.
    */
  def insertRight(index: Position, value: A, newPos: Position): RGArray[A]
  /** 
    * Removes an element stored under provided index. This is a soft delete,
    * an element will still occupy memory until a removeTombsones will be called.
    */
  def delete(index: Position): RGArray[A]
  /** Returns a new RGArray without elements marked as deleted. */
  def removeTombstones(): RGArray[A]
  /** Return a Vertex for an index value relative to current state of an array */
  def getVertex(index: Int): Vertex[A]
}
```


krasserm:
@Horusiath thanks a lot for your contribution! That's a great addition to have in Eventuate. 

Regarding your question:

> This is the part I'm struggling with, as I don't know how to expose current emitterId from CRDTService level.

[CRDTService.effect](https://github.com/RBMHTechnology/eventuate/blob/4d6624a15ff6a923aac731babe1757e913ab77e0/eventuate-crdt/src/main/scala/com/rbmhtechnology/eventuate/crdt/CRDTService.scala#L66) has an `event` parameter from which you can obtain the `emitterId`.

I'll take a closer look and comment on your commit(s) next week.

krasserm:
@Horusiath I'll defer review until CI passes. Let me know if you have any question in the meantime. Thanks!

Horusiath:
@krasserm I haven't implemented protobuf serialization. What command/settings should I call to regenerate protobuf classes from proto file?

krasserm:
They are generated automatically with compile.

krasserm:
I just found the [RGArray tests failed](https://travis-ci.org/RBMHTechnology/eventuate/jobs/280468023#L4575) although Travis is reports success - sorry for the inconvenience. I'll check the CI setup later.

I can also reproduce the test errors locally. In sbt:

    > project crdt
    > test:testOnly

Horusiath:
This should be ready to review by now. Also there is one design issue (I've mentioned it on gitter). At this moment element insertion is dependent explicitly on vertex position identifier i.e:

```scala
val emitterId = "a"
RGArray.empty[String]
  .insertRight("Hello", Position(1, emitterId))
  .insertRight(Position(1, emitterId), "world", Position(2, emitterId))
```

However this is a pretty obscure API. What I'd like to achieve is something like:

```scala
implicit val emitterId = "a"
RGArray.empty[String]
  .insertRight(0, "Hello")
  .insertRight(1, "world")
```

It's achievable by getting a `Position` identifier based on an index of a resolved `RGArray[A].value` and by passing emitterId implicitly. 

This can work for `RGArray[A]`, however eventuate also brings the convention of operations defined in corresponding `CRDTService`. This is when I've got a problem, as I'd need to emit an operation in `prepare` phase based on a state of the CRDT at the point when that operation was called (in order to get `Position` based on a value index). I'll need help here.

krasserm:
> I'll need help here.

Does [that answer](https://gitter.im/RBMHTechnology/eventuate?at=59d717e2614889d4757f4d72) help?

Horusiath:
@krasserm I think, this one is ready for review now.

I've dropped the idea with exposing insert operation as `RGArray[A].insertRight(index: Int, value: A)`. This is because RGArray inserts requires using absolute positions, while Int-based index is relative to a current state of the RGArray.

Therefore if one would, like to insert a value at some index, he'd have to retrieve CRDT instance first, and then retrieve absolute position for that index using `RGAarray[A].getVertex(index:Int).pos`.

Example:

```scala
val service = new RGArrayService[String](s"service-$locationId", endpoint.logs(logName))
val crdtId = "1"
import serviceA.system.dispatcher

for {
  rga <- service.insertRight(crdtId, Position.head, "hello")
  position = rga.getVertex(0).pos  // integers are state-relative, we need absolute position id
  rga2 <- service.insertRight(crdtId, position, "world")
} yield rga2.value  // Vector("hello", "world")
```

Horusiath:
I've found one feature that still should be implemented.

## Automatic garbage collection of tombstoned values

`RGArray` allows to remove elements - however it does so by tombstoning. This means that tombstoned elements have to be collected from time to time.

Current implementation doesn't have an automatic support for it, it must be done manually. I think, I know how to make GC work automatically:

1. When user wants to remove indexed element from RGArray, we mark this element with tombstone `Some[VectorTime]`- its value is the latest local timestamp for that CRDT (even persisted yet). 
2. When we confirm a replication event with some `VectorTime`, we can compute the latest stable timestamp - a highest `VectorTime` which says at which casual point in time all endpoints have persisted their events. This way we know, that all events in its casual past have been persisted and delivered. Therefore we can remove all tombstoned elements that happened before that point, because all replicas acknowledged them as removed and they are no longer necessary.

I'm still looking for a way to support this in eventuate.

krasserm:
@Horusiath Can you please share a link to the formal specification you used for implementing `RGArray`?

Horusiath:
@krasserm A formal specification can be found here (Specification 19): http://hal.upmc.fr/file/index/docid/555588/filename/techreport.pdf

Some points:

- This specification does not describe the problem of pruning tombstones.
- Timestamps are described as tuple *(local-clock, client-UID)*. For eventuate I've used logical monotonic  clock (current counter is stored inside RGArray itself) and emitterId instead of client-UID.
- Spec describes RGArray as a variant of 2P-Set. I've used `Vector[Vertex[A]]` instead with vertex tombstone marking. I've did it for few reasons:
  - Building output `Vector[A]` (which for me seems to be a desired Scala collection value for `RGArray[A]`) with correctly ordered elements is trivial.
  - Mapping output `Vector[A]` index to a vertex position is also pretty simple.
  - Since tombstone marker is a field (rather than separate collection) I can use it to carry some extra metadata, that can be later used for tombstone pruning.

I've found more detailed study [here](https://hal.inria.fr/hal-01343941/document), but I haven't tried to implement this approach yet.

eventuate-crdt/src/main/scala/com/rbmhtechnology/eventuate/crdt/RGArray.scala
```diff
@@ -0,0 +1,165 @@
...
+      }
+    })
+    this.copy(vertices = pruned)
+  }
+
+  @tailrec
+  private def skipOver(pos: Position, idx: Int): Int = {
+    if (idx < vertices.length && vertices(idx).pos > pos) skipOver(pos, idx + 1) else idx
+  }
+}
```
krasserm: Please review parameter naming in `RGArray` class:

- parameters with `pos` in name have type `Position` (expected) or `Int` (not expected)
- parameters of type `Position` have names containing `pos` (expected), `id` (not expected) or `index` (not expected)
- ...

I skip listing other examples, hope it's clear what I mean.

eventuate-crdt/src/main/scala/com/rbmhtechnology/eventuate/crdt/CRDTSerializer.scala
```diff
@@ -137,6 +162,34 @@ class CRDTSerializer(system: ExtendedActorSystem) extends Serializer {
...
+
+  private def rgArrayVertexFormatBuilder(vertex: Vertex[_]): RGArrayVertexFormat.Builder = {
+    val builder = RGArrayVertexFormat.newBuilder
+      .setPosition(rgArrayPositionFormatBuilder(vertex.pos))
+      .setValue(payloadSerializer.payloadFormatBuilder(vertex.value.asInstanceOf[AnyRef]))
+
+    vertex.tombstone match {
+      case Some(timestamp) => builder.setTombstone(commonSerializer.vectorTimeFormatBuilder(timestamp))
+      case None            => ()
+    }
```
krasserm: `vertex.tombstone.foreach { ... }`

eventuate-crdt/src/main/protobuf/CRDTFormats.proto
```diff
@@ -32,6 +32,22 @@ message ORCartEntryFormat {
   optional int32 quantity = 2;
 }
 
+message PositionFormat {
+  optional int32 order = 1;
```
krasserm: Rename to `index`?

eventuate-crdt/src/main/scala/com/rbmhtechnology/eventuate/crdt/RGArray.scala
```diff
@@ -0,0 +1,165 @@
...
+
+object Position {
+  val head: Position = Position(0, "")
+}
+
+/**
+ * Bucket for values stored inside an [[RGArray]].
+ * @param value Assigned value.
+ * @param pos Position of that value in relation to others within [[RGArray]].
+ * @param tombstone Determines a local casual time, at which current value has been removed.
```
krasserm: Rename to `deletionTime`? with documentation like *Determines the vector time at which this value has been removed*?

eventuate-crdt/src/main/protobuf/CRDTFormats.proto
```diff
@@ -59,4 +75,14 @@ message AddOpFormat {
 message RemoveOpFormat {
   optional PayloadFormat entry = 1;
   repeated VectorTimeFormat timestamps = 2;
+}
+
+message InsertOpFormat {
+  optional PayloadFormat value = 1;
+  optional int32 pos = 2;
```
krasserm: Maybe this can be removed if we can re-use local sequence numbers.

eventuate-adapter-spark/src/main/scala/com/rbmhtechnology/eventuate/adapter/spark/SparkStreamAdapter.scala
```diff
@@ -235,4 +235,3 @@ private class DurableEventReceiver(id: String, connection: ReplicationConnection
       readSchedule.foreach(_.cancel())
   }
 }
-
```
krasserm: Please undo all changes to files that are not related to the features of this PR.

eventuate-crdt/src/it/scala/com/rbmhtechnology/eventuate/crdt/CRDTSerializerSpec.scala
```diff
@@ -93,12 +118,24 @@ class CRDTSerializerSpec extends WordSpec with Matchers with BeforeAndAfterAll {
 
       serialization.deserialize(serialization.serialize(initial).get, classOf[ORCart[_]]).get should be(expected)
     }
+    "support RGArray serialization with default key serialization" in serializations.tail.foreach { serialization =>
```
krasserm: Please rename test, as it causes a [DuplicateNameException](https://travis-ci.org/RBMHTechnology/eventuate/jobs/291251606#L4026) (sorry, build issues still not fixed).

krasserm: After I fixed the duplicate name error locally and ran `CRDTSerializerSpec`, all `RGArray`-related tests failed. Maybe it's a good idea to run tests locally before pushing to Github e.g. with `it:testOnly com.rbmhtechnology.eventuate.crdt.CRDTSerializerSpec`.

eventuate-crdt/src/main/scala/com/rbmhtechnology/eventuate/crdt/RGArray.scala
```diff
@@ -29,46 +29,45 @@ import scala.util.{ Failure, Success, Try }
...
    */
-  def delete(index: Position, time: VectorTime) = {
-    this.copy(vertices = vertices.map((v: Vertex[A]) => if (v.pos == index) v.copy(tombstone = Some(time)) else v))
+  def delete(pos: Position, time: VectorTime) = {
+    this.copy(vertices = vertices.map((v: Vertex[A]) => if (v.pos == pos) v.copy(deletionTime = Some(time)) else v))
   }
 
   /**
-   * Prunes all removed values, that have been tombstoned in `timestamp`s casual past.
+   * Prunes all removed values, that have been tombstoned in `timestamp`s causual past.
```
krasserm: causal

Horusiath:
I think, I'll close this one for now. So far my conclusion is that efficient `RGArray[A]` should be a mutable data structure, and this doesn't work well with existing eventuate design. All immutable versions I can think off can produce O(N) allocations, potentially rebuilding a whole structure on write, or need to build a linear sequence of data directly by tree hops, which is O(log(N) * N) on reads.

krasserm:
See my comment at https://gitter.im/RBMHTechnology/eventuate?at=5a4dcb03b48e8c35669a966d

PR #400, opened 2017-11-08
Fix event-loss with processors and replication

### LINKED ISSUE #399: Using EventsourceProcessors in combination with Replication may lead to loss of processed events
The implementation of an EventsourcedProcessor is based on the replication protocol and thus it applies the causality filter to processed events, however the causality filter is not based on the final vector timestamp the processed event should get, but on the vector timestamp of the source event. In combination with replication of source and target log of the processor this may lead to losing (i.e. filtering out) the processed event.

For example:
Given two location A and B with two replicated logs L1 and L2 and each with a processor processing from L1 to L2 only locally emitted events (to avoid duplicate processing of events). Vector timestamps have the four process ids: A_L1, A_L2, B_L1, B_L2 and are denoted in the form (1,2,3,4)
- A emits e1 (1,0,0,0) to L1
- B replicates e1 to L1
- B emits e2 (1,0,2,0) to L1
- B processes e2 to L2: e3 (1,0,2,1)
- A replicates e3 to L2
- A processes e1 to L2: e4 (1,2,0,0)

Problem is that e4 is initially created with timestamp of its source event (1,0,0,0) then filtered by the causality filter according to the version vector of the target log L2 which is (1,0,2,1) in this example and thus filtered out before its timestamp is finalized to (1,2,0,0) (which would not be filtered out)

volkerstampa:
To avoid that events emitted by processors are filtered by the causality
filter of the utilized replication implementation, events that have an
undefined process id are considered incompletely initialized and thus
are never smaller than a given vector time.

Closes #399

### DIFF
--- a/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/DurableEvent.scala
+++ b/eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/DurableEvent.scala
@@ -93,9 +93,12 @@ case class DurableEvent(
 
   /**
    * Returns `true` if this event happened before or at the given `vectorTime`.
+   *
+   * An event with undefined [[processId]] has an incomplete initialized [[vectorTimestamp]] and is
+   * therefore considered to be never before a given [[VectorTime]].
    */
   def before(vectorTime: VectorTime): Boolean =
-    vectorTimestamp <= vectorTime
+    processId != UndefinedLogId && vectorTimestamp <= vectorTime
 
   /**
    * The default routing destination of this event is its `emitterAggregateId`. If defined, the event is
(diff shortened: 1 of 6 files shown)


magro:
LGTM

eventuate-core/src/main/scala/com/rbmhtechnology/eventuate/DurableEvent.scala
```diff
@@ -93,9 +93,12 @@ case class DurableEvent(
 
   /**
    * Returns `true` if this event happened before or at the given `vectorTime`.
+   *
+   * An event with undefined [[processId]] has an incomplete initialized [[vectorTimestamp]] and is
+   * therefore considered to be never before a given [[VectorTime]].
    */
   def before(vectorTime: VectorTime): Boolean =
-    vectorTimestamp <= vectorTime
+    processId != UndefinedLogId && vectorTimestamp <= vectorTime
```
krasserm: This breaks idempotency of event processing.

volkerstampa:
The _fix_ destroys the idempotency guarantee of an EventsourcedProcessorand is therefore declined.

PR #404, opened 2018-02-01
Skip empty partitions on replay from Cassandra

### LINKED ISSUE #403: Replay from Cassandra stops prematurely if there are large sequence number gaps covering empty partitions
After disaster recovery the local sequence is potentially adjusted to ensure that it is larger than the local log's time (see also [commit](https://github.com/RBMHTechnology/eventuate/pull/308/commits) for #286). If Cassandra is used as backend and the gap introduced through this adjustment covers empty partitions, replays from this event log stop prematurely as they do to proceed after the first empty partition.

volkerstampa:
Do not stop reading events from EventLog on empty partitions if
the toSequenceNr is known (i.e. not Long.MaxValue)

Closes #403


magro:
FWIW, LGTM!

